Issue #132, opened 2017-07-22
Creating an all day event on August 1 results in an all day event July 31


jegjessing:
I added an event and set start and end at August 1, but the event appear in July 31.
I'm on Europe/Copenhagen region, and I try to add the event in the Google calender using Maya

welaq:
Yes, same issue here even when trying to add the event into local calendar.

PR #136, opened 2017-08-19
Correctly generate date ranges for all day events (#132)

Thalexander:
Adding the interval can cause in certain timezones all day events to show up on the wrong day, particularly the day before.

### DIFF
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -156,9 +156,6 @@ namespace Maya.Util {
         bool allday = is_all_day (start, end);
         if (allday) {
             end = end.add_days (-1);
-            var interval = (new DateTime.now_local ()).get_utc_offset ();
-            start = start.add (-interval);
-            end = end.add (-interval);
         }
 
         start = strip_time (start);


matzipan:
I'm merging this one in since I don't think it would cause any harm.

renearts:
Looks like the same bug as this one: https://bugs.launchpad.net/maya/+bug/1421012
I guess your contribution might fix these issues as well (there are some more launchpad reports); if so, don't forget to collect the bounty: https://www.bountysource.com/issues/8647856-events-wrongly-displayed-on-two-days

matzipan:
Thanks @renearts. I tried to find the old bug report but I couldn't find it.

renearts:
I see you were able to comment on it, what do you mean? All links work fine here..

matzipan:
No, I meant I tried before you commented but I couldn't find it.

renearts:
Ah great, glad I was able to help. Thanks for fixing this annoying bug!

Thalexander:
This should only affect all day events so I don't believe it is related to that bug, however it may be related to #9

Issue #137, opened 2017-09-03
Agenda view doesn't show multi-day all day events correctly


Thalexander:
When creating an all-day event, if the event extends over multiple days the agenda view will not show it on the last day in certain timezones

matzipan:
I can confirm the bug

PR #138, opened 2017-09-03
Show multi-day all day events correctly #137

Thalexander:
Similarly to #136 the interval being added can cause the agenda view to show the event on the wrong days

### DIFF
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -119,9 +119,6 @@ namespace Maya.Util {
         bool allday = is_all_day (start, end);
         if (allday) {
             end = end.add_days (-1);
-            var interval = (new DateTime.now_local ()).get_utc_offset ();
-            start = start.add (-interval);
-            end = end.add (-interval);  
         }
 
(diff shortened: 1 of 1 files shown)


jeremypw:
Please describe the bug being fixed more clearly with an example for testing.  When testing this PR, I was still getting incorrect date ranges for multi-day all-day events.  Used 2 day events in a European timezone, both within and across month boundaries (September/October).  In the right-hand panel (Today's events and Upcoming events) these events were shown as 3 day events, ending on the day following the correct one.

matzipan:
This seems linked to bug report #137

jeremypw:
If I have understood the bug correctly, this PR does not seem to fix it for me.

matzipan:
@jeremypw to test this create an all-day event spanning multiple days and then in month view click on the last of the days that that event should span and in master, the agenda view will not show the event.

jeremypw:
OK, I can confirm that this bug is solved.  It appears I was looking at a different, related bug that is not fixed.

PR #139, opened 2017-09-04
Implement exporting of calendar #30

### LINKED ISSUE #30: Exporting calendar doesn't work
When you try to export your calendar to .ics (via contextual menu) nothing happened.

When you select "export the calendar..." You can select where you want to save your file but no files are created.

Maybe this bug it's linked to Bug #1091482 (and saying the import function isn't implemented yet) Maybe it's the same thing for export function ?

Launchpad Details: [#LP1355090](https://bugs.launchpad.net/bugs/1355090) Unknown - 2014-08-11 09:40:29 +0000

### LINKED ISSUE #80: "Send by email" is not working properly
In case of using "Send by email" feature on the top menu, I think it would probably do first file or zip schedules or events as a file.
and then attach it to the new mail, but it doesn't seem to be worked like this.
please see the screenshot.

Launchpad Details: [#LP1557485](https://bugs.launchpad.net/bugs/1557485) Jung-Kyu Park - 2016-03-15 11:59:08 +0000

### LINKED ISSUE #129: The print function yields a blank page
The print function shows a blank page in the preview, and only prints a blank page.

Thalexander:
Whilst the user interface shows an export button, the function to export was not implemented, these commits will add this and as such all of the contractors function now (print (#129), compress, email (#80), bluetooth, export). Print just prints the text, however that seems to be the expected behaviour based on trying to print ics files from Files. The exported files may contain some errors namely about having no values for certain fields, however this does not impact usage of the ical file.

Fixes #80 
Fixes #129
Fixes #30

### DIFF
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -709,7 +709,6 @@ namespace Maya.Util {
             builder.append (event.get_as_string ());
         }
         builder.append ("END:VCALENDAR");
-        string output = builder.str;   
 
         string file_path = GLib.Environment.get_tmp_dir () + "/calendar.ics";
         try {
@@ -718,7 +717,7 @@ namespace Maya.Util {
                 file.delete ();
             }
             var dos = new DataOutputStream (file.create (FileCreateFlags.REPLACE_DESTINATION));
-            uint8[] data = output.data;
+            uint8[] data = builder.data;
             long written = 0;
             while (written < data.length) {
                 written += dos.write (data[written:data.length]);
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -699,8 +699,33 @@ namespace Maya.Util {
      * ical Exportation
      */
 
-    public void save_temp_selected_calendars (){
-        //TODO:Create the code !
+    public void save_temp_selected_calendars () {
+        var calmodel = Model.CalendarModel.get_default ();
+        var events = calmodel.get_events ();
+        var builder = new StringBuilder ();
+        builder.append ("BEGIN:VCALENDAR\n");
+        builder.append ("VERSION:2.0\n");
+        foreach (E.CalComponent event in events) {
+            builder.append (event.get_as_string ());
+        }
+        builder.append ("END:VCALENDAR");
+        string output = builder.str;   
+
+        string file_path = GLib.Environment.get_tmp_dir () + "/calendar.ics";
+        try {
+            var file = File.new_for_path (file_path);
+            if (file.query_exists ()) {
+                file.delete ();
+            }
+            var dos = new DataOutputStream (file.create (FileCreateFlags.REPLACE_DESTINATION));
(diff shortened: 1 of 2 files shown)


core/Utils.vala
```diff
@@ -699,8 +699,33 @@ namespace Maya.Util {
...
+        var calmodel = Model.CalendarModel.get_default ();
+        var events = calmodel.get_events ();
+        var builder = new StringBuilder ();
+        builder.append ("BEGIN:VCALENDAR\n");
+        builder.append ("VERSION:2.0\n");
+        foreach (E.CalComponent event in events) {
+            builder.append (event.get_as_string ());
+        }
+        builder.append ("END:VCALENDAR");
+        string output = builder.str;
```
jeremypw: This is not required; you can get the data directory from the StringBuilder with uint8[] data =  builder.data;

core/Utils.vala
```diff
@@ -699,8 +699,33 @@ namespace Maya.Util {
...
+        string output = builder.str;   
+
+        string file_path = GLib.Environment.get_tmp_dir () + "/calendar.ics";
+        try {
+            var file = File.new_for_path (file_path);
+            if (file.query_exists ()) {
+                file.delete ();
+            }
+            var dos = new DataOutputStream (file.create (FileCreateFlags.REPLACE_DESTINATION));
+            uint8[] data = output.data;
```
jeremypw: Get data directly from builder (see above).

core/Utils.vala
```diff
@@ -699,8 +699,33 @@ namespace Maya.Util {
...
+            long written = 0;
+            while (written < data.length) {
+                written += dos.write (data[written:data.length]);
+            }
+        } catch (Error e) {
+            warning ("%s\n", e.message);
+        }
     }
```
jeremypw: Wouldn't be simpler to create an empty file and the use FileUtils.set_contents (filename, string, -1)?

core/Utils.vala
Thalexander: Thanks, I have now switched to using this.

core/Utils.vala
Thalexander: There is very limited documentation of this on valadoc, although there is a fixed bug on Scratch (https://bugs.launchpad.net/scratch/+bug/1021860) that gives some reasons it may be undesirable. I think my main concern with that would be whether it ensures all data is written before returning as the file may be immediately moved by the contractor (export) which could cause issues if the file isn't fully written, which is why I didn't just write the string using the dos. I was unaware of FileUtils so I am not sure of how it works, and whether it ensures the files are written, particularly with the documentation.

jeremypw: Agreed, as suggested by that bug report you should use file.replace_contents () rather than FileUtils.set_contents (). If you use the synchronous version then your concerns should be addressed.

PR #142, opened 2017-09-14
Correct date ranges in Agenda

### LINKED ISSUE #141: All-day events over multiple days show incorrect date range in agenda pane.
If an event covering more than one day is created with the "all day event" switch "on", then the agenda shows an incorrect date range for this event.  The last day is shown one day later than the true last day.
screenshot from 2017-09-14 11 13 51

Thalexander:
Handles all day events correctly when getting local datetimes from ical components

Fixes #141

### DIFF
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -110,6 +110,11 @@ namespace Maya.Util {
 
         start_date = Util.ical_to_date_time (dt_start);
         end_date = Util.ical_to_date_time (dt_end);
+
+        bool allday = is_all_day (start_date, end_date);
+        if (allday) {
+            end_date = end_date.add_days (-1);
+        }
     }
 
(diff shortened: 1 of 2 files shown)


matzipan:
Agreed with Jeremy.

Thalexander:
A day is only taken away from the end date, I do not believe this would be possible to handle in a sensible way in ical_to_date_time as that handles any date component.

matzipan:
Hmm you're right, I thought this date was originating from extracting the date wrongly. Any clue why the ical standard is defined this way?

Thalexander:
The range is defined to be exclusive of the end date ("The "DTEND" property for a "VEVENT" calendar component specifies the non-inclusive end of the event"), quite a few languages use an inclusive start exclusive end for slicing arrays too so it's not too unusual.

PR #146, opened 2017-09-15
Grab focus on event title when creating event

### LINKED ISSUE #48: No focus for event title when creating an event
When double-clicking on a day, the "Create new event" dialogue opens. Intuitively I would just begin to type the event name, but the "Title" text box has no focus. So you have to click on it first to start typing. When doing so and you start typing, the text box loses focus again and you have to click it again to continue.

Launchpad Details: [#LP1421961](https://bugs.launchpad.net/bugs/1421961) greenshift - 2015-02-14 15:09:30 +0000

Thalexander:
When creating an event it would be natural to grab focus to the title entry, note this will also grab focus and select the existing title when editing an event too, this behaviour can be changed quite easily and is really a design choice, I quite like it as it doesn't leave the keyboard unfocused when editing an event, however let me know your thoughts on it. 

Fixes #48


matzipan:
Works as advertised, however I'm unsure about hooking this to `draw`. I tried using other signals or simply doing this after `show_all` but it didn't work, which is why I added Jeremy and Adam as reviewers.

matzipan:
I've tried to change it myself with realize but I couldn't get it to work. What Alex is doing seems to be recommended [here](http://lethalman.blogspot.fr/2012/11/grab-focus-on-gtk-widget.html)

jeremypw:
Apologies - using "realize" causes the entry text to be selected (blue background) and I was assuming it also had keyboard focus.  Using "draw" still seems hacky (even using "realize" was not ideal, although simple) - there must be a better way ...  I'll experiment some more.

jeremypw:
This works:
```
 realize.connect (() => {
     Idle.add (() => {
        title_entry.grab_focus ();
    });
 });

```

but is hardly any better than the original.

jeremypw:
I have not found any more elegant solution, so I withdraw my objection to the original solution.

Issue #53, opened 2017-04-17
Set location by double-clicking on the map


elementaryBot:
Should it be possible to set the location of an event by double-clicking on the map? I wasn't sure if the inability to do so was a bug or if I had misunderstood the purpose of the map.

Launchpad Details: [#LP1445890](https://bugs.launchpad.net/bugs/1445890) Corentin Noël - 2015-04-19 10:08:50 +0000

elementaryBot:
Double clicking don't seems to be the way this should be implemented.
If it's possible this could be done by dragging the location marker. 

Launchpad Details: [#LPC](https://bugs.launchpad.net/maya/+bug/1445890/comments/1) Steffen Schuhmann - 2015-05-28 06:43:19 +0000

danrabbit:
Scrolling is already available for zoom and I think is the more standard/accepted way to zoom

jeremypw:
Be that as it may, I cannot see any downside other than a minimal amount of extra code.

danirabbit:
I'm going to close this as "Design conflict". No other calendar app that I can see lets you select location by interacting with the map. The map is read-only and location has to be set through the location entry

PR #147, opened 2017-09-16
Allow setting location using map

Thalexander:
The map currently only displays the location found from the search bar and does not allow setting location through it. This problem was raised by issue #53; as double clicking is used to zoom, I agree with Steffen that the logical method is through dragging. This push will implement this, however note this will not use the exact way which would be through latitude and longitude but rather reverse geocodes to get an address as I think that is a better way of representing the location to the user, even if it loses some accuracy.

### DIFF
--- a/src/EventEdition/LocationPanel.vala
+++ b/src/EventEdition/LocationPanel.vala
@@ -201,6 +201,7 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
         find_cancellable = new GLib.Cancellable ();
         Geocode.Location location = new Geocode.Location (latitude, longitude);
         var reverse = new Geocode.Reverse.for_location (location);
+        
         try {
             var address = yield reverse.resolve_async (find_cancellable);
             var builder = new StringBuilder ();
--- a/src/EventEdition/LocationPanel.vala
+++ b/src/EventEdition/LocationPanel.vala
@@ -207,34 +207,28 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
             var builder = new StringBuilder ();
             if (address.street != null) {
                 builder.append (address.street);
-                if (address.town != null) {
-                    builder.append (", ");
-                    builder.append (address.town);
-                }
-                if (address.county != null) {
-                    builder.append (", ");
-                    builder.append (address.county);
-                }
-                if (address.postal_code != null) {
-                    builder.append (", ");
-                    builder.append (address.postal_code);
-                }
-                if (address.country != null) {
-                    builder.append (", ");
-                    builder.append (address.country);
-                }
+                add_address_line (builder, address.town);
+                add_address_line (builder, address.county);
+                add_address_line (builder, address.postal_code);
+                add_address_line (builder, address.country);
             } else {
                 builder.append (address.name);
-                if (address.country != null) {
-                    builder.append (", ");
-                    builder.append (address.country);
-                }
+                add_address_line (builder, address.country);
             }
+            
             location_entry.text = builder.str;
         } catch (Error error) {
             debug (error.message);
         }
     }
+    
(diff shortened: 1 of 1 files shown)


Thalexander:
It looks like the build failed to fetch the dependencies... Is there a way to make it try again?

src/EventEdition/LocationPanel.vala
```diff
@@ -187,6 +189,46 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
             debug (error.message);
         }
     }
+
+    private async void find_location (double latitude, double longitude) {
+        if (find_cancellable != null)
+            find_cancellable.cancel ();
```
jeremypw: elementaryos code style: Single line if clauses should still have braces.  Also should be followed by blank line.

src/EventEdition/LocationPanel.vala
```diff
@@ -187,6 +189,46 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
...
+                if (address.postal_code != null) {
+                    builder.append (", ");
+                    builder.append (address.postal_code);
+                }
+                if (address.country != null) {
+                    builder.append (", ");
+                    builder.append (address.country);
+                }
+            }
+            else {
```
jeremypw: elementaryos code style:  else and braces should be on same line:  `} else {`

src/EventEdition/LocationPanel.vala
```diff
@@ -187,6 +189,46 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
             debug (error.message);
         }
     }
+
+    private async void find_location (double latitude, double longitude) {
+        if (find_cancellable != null)
+            find_cancellable.cancel ();
+        Geocode.Location location = new Geocode.Location (latitude, longitude);
+        var reverse = new Geocode.Reverse.for_location (location);
+        try {
```
jeremypw: I would prefer a preceding blank line for readability

src/EventEdition/LocationPanel.vala
jeremypw: As far as I can see, find_cancellable is never created?  It is recommended to create a new cancellable when required rather than re-using the old one.

src/EventEdition/LocationPanel.vala
Thalexander: Where in particular, before the try?

jeremypw: Yes, that's right:

```
       var reverse = new Geocode.Reverse.for_location (location);

       try {
           ...
```

src/EventEdition/LocationPanel.vala
```diff
@@ -187,6 +192,49 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
...
+                    builder.append (address.county);
+                }
+                if (address.postal_code != null) {
+                    builder.append (", ");
+                    builder.append (address.postal_code);
+                }
+                if (address.country != null) {
+                    builder.append (", ");
+                    builder.append (address.country);
+                }
```
jeremypw: You could avoid repeating code by using a small function to add address components e.g.
```
private void add_address_line (StringBuilder sb, string? text) {
    if (text != null) {
         sb.append (", ");
         sb.append (text);
    }
}

File: src/EventEdition/LocationPanel.vala [hunk 5]
```diff
@@ -187,6 +192,49 @@ public class Maya.View.EventEdition.LocationPanel : Gtk.Grid {
...
+            } else {
+                builder.append (address.name);
+                if (address.country != null) {
+                    builder.append (", ");
+                    builder.append (address.country);
+                }
+            }
+            location_entry.text = builder.str;
```
jeremypw: elementary code style:  Should be blank line after end of last if clause (before location_entry.text = ...)

Issue #42, opened 2017-04-17
Location Text broken in Popup


elementaryBot:
installed as beta1, updated to newest packages.

the calendar is CalDAV from Zimbra 8.

it is a repeated event. location "Technik" get's correctly displayed in the sidebar to the right but not in the popup. (see screenshot)

i started maya in the terminal and get this logs:

davidak@elementary-freya-4:~$ maya-calendar 

(process:5195): Gtk-WARNING **: Locale not supported by C library.
	Using the fallback 'C' locale.

(maya-calendar:5195): Clutter-WARNING **: Locale not supported by C library.
Using the fallback 'C' locale.
[_LOG_LEVEL_INFO 16:24:53.830705] Application.vala:155: Calendar version: 0.3
[_LOG_LEVEL_INFO 16:24:53.830811] Application.vala:157: Kernel version: 3.13.0-43-generic
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
[_LOG_LEVEL_WARN 16:24:57.794143] RepeatPanel.vala:319: 3
[_LOG_LEVEL_WARN 16:25:03.316082] RepeatPanel.vala:319: 3
[_LOG_LEVEL_FATAL 16:26:01.908242] geocode_forward_add: assertion 'value == NULL || g_utf8_validate (value, -1, NULL)' failed
[_LOG_LEVEL_FATAL 16:26:01.908352] Calendar will not function properly.
[_LOG_LEVEL_WARN 16:26:01.914453] RepeatPanel.vala:319: 3
[_LOG_LEVEL_WARN 16:26:01.935941] [Pango] Invalid UTF-8 string passed to pango_layout_set_text()
[_LOG_LEVEL_WARN 16:26:04.673667] [Pango] Invalid UTF-8 string passed to pango_layout_set_text()

(apport-gtk:5279): Gtk-WARNING **: Locale not supported by C library.
	Using the fallback 'C' locale.
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug

*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug

*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug

[...]

PR #152, opened 2017-09-21
Avoid garbled text and crash in Location Panel.

jeremypw:
see #42 

TO TEST:

1) Open an event with a location set; open Location tab.
2) Cancel
3) Repeat 1) and 2)

In trunk the location entry text becomes garbled and the app may crash.
In branch the entry text is correct each time.


jeremypw:
Marked as high priority as the app may crash and there is text corruption.

PR #164, opened 2017-09-24
Use Gtk.StackSwitcher in EventDialog

danrabbit:
Replaces the modebutton with a stack switcher and some other minor cleanup while I'm here. I let the stack switcher fill just to have bigger click targets because we have room to do it

screenshot from 2017-09-23 17 56 52

To be clear, this should be tested on Loki to make sure that the icon thing works there. I'm not sure when that was added


danrabbit:
Yeah it needs to be fixed but I didn’t want to break translations

PR #167, opened 2017-09-28
Install app icons to hicolor

### LINKED ISSUE #103: don't reference and expect non-standard icons, ship branded icons instead
Most elementary desktop apps don't ship the icons (to /usr/share/icons/hicolor) they reference in their .desktop files. This results in some problems (especially on other linux distributions, but also on other DEs), including:

1) Apps not shipping their own icon might show in Application Menus with the "missing icon" icon.

2) The affected apps will not show up in "Application Centers" which are using appdata for metadata, because the metadata is deemed to be invalid due to a non-existent icon (not in the package) being referenced at metadata generation time.

This behavior is intentional, because metadata generation for one package cannot (and should not) depend on other packages being present (to be reproducible).

For example, some elementary apps don't show up in GNOME Software because of that - this affects at least ubuntu and fedora, probably other distros too.

The affected packages are (I may have missed some in that list):

application          icon name
--------------------------------------------
audience             multimedia-video-player
maya-calendar        office-calendar
screenshot-tool      accessories-screenshot
snap-photobooth      accessories-camera
wingpanel            wingpanel

Launchpad Details: [#LP1658325](https://bugs.launchpad.net/bugs/1658325) Fabio Valentini - 2017-01-21 14:29:38 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44229538-don-t-reference-and-expect-non-standard-icons-ship-branded-icons-instead?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Thalexander:
This takes the office-calendar icons from elementary icons and includes it in the project and installs it (following the new rddn scheme for the icon name) to hicolor.

Fixes #103


jeremypw:
What are the implications of having differently named icons in the elementary icon theme?  Shouldn't the theme icon be used if present, falling back to hicolor if not?

Thalexander:
Well as it in non-standard, when installed on other distros the icon will not exist, and it prevents it showing up in some app stores (as detailed by #103.)

jeremypw:
So, are you suggesting that the icon should not be in the elementary theme at all?  (Another distro could have the elementary icon theme installed).

Thalexander:
As it is not part of the spec, the icon will only apply to elementary calendar so yes it makes sense to ship the icon with the app rather than in the icon pack.

jeremypw:
As there are wider implications I'll leave to the UI team to approve/schedule this PR.  I cannot see anything obviously amiss in the code 'though.

cassidyjames:
Since there are conflicts and there has been no activity for over a year, I'm going to close this PR. Feel free to update it and we can reopen it, though. I've also asked for clarification in Slack whether or not we want to ship the app icons in the repo.

PR #171, opened 2017-10-01
Sort events on Grid

### LINKED ISSUE #94: Events not time sorted in calendar
Events are randomly sorted in calendar's day squares although they are properly time sorted when shown on the right panel.

Launchpad Details: [#LP1629140](https://bugs.launchpad.net/bugs/1629140) Josh - 2016-09-29 23:41:54 +0000

Thalexander:
This sorts the events on the grid according to the Util.compare_events function (it moves non-EventButtons to the end as the only one of those should be the show more label) .

Fixes #94

### DIFF
--- a/src/Grid/EventButton.vala
+++ b/src/Grid/EventButton.vala
@@ -104,15 +104,4 @@ public class Maya.View.EventButton : Gtk.Revealer {
         rgba.parse (color);
         event_box.override_background_color (Gtk.StateFlags.NORMAL, rgba);
     }
-
-    /**
-     * Compares the given buttons according to date.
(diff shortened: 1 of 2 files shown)


tintou:
Wouldn't it be better to rewrite the grid with a Gtk.ListBox ? It has automatic sorting functions

Thalexander:
There would be minimal reductions in the code size (about 4 lines) and then as the Widgets would be packaged in GtkListBoxRows the children of these rows would need to be iterated though to hide or reveal them, which would in my view negate the performance and complexity advantages introduced by the sorting.

PR #172, opened 2017-10-03
Follow week setting in System Settings

### LINKED ISSUE #170: Inconsistent handling of settings
The Date & Time section of the System Settings app has an option to show week numbers, however we ignore this setting and follow the saved state of the application and allow hiding it by right clicking on the days, I would argue this behaviour is not very discoverable as it relies on users right clicking on the week days.

Despite us following the saved state here, we do follow the AM,PM/24h setting in Date & Time so we inconsistently follow the settings, I think we should just follow the setting in Date & Time for easier discoverability and more consistency.

However I concede potentially people may want different settings for wingpanel and Calendar in which case maybe this is an issue with the settings app and it should be moved to the date time indicator.

Thalexander:
This changes the Week Settings to follow "org.pantheon.desktop.wingpanel.indicators.datetime" rather than "org.pantheon.maya.savedstate", whilst the domains would suggest this to not make sense, to users they are only shown a "Show week numbers" toggle in System Settings and the method for hiding week numbers in Calendar is not very discoverable. I am happy with you not accepting this as a design decision to favour handling the settings separately; I do think the setting should be clarified to be only for wingpanel in System Settings if that is the case.

Fixes #170


Thalexander:
It now lookup if the key exists and if it doesn't falls back to the SavedState settings, also it has been rewritten to use GLib.Settings.

core/Utils.vala
```diff
@@ -720,5 +720,29 @@ namespace Maya.Util {
     public string get_hexa_color (Gdk.RGBA color) {
         return "#%02X%02X%02X".printf ((uint)(color.red*255), (uint)(color.green*255), (uint)(color.blue*255));
     }
+    
+    /*
+     * Handles Week Settings
+     */
```
danrabbit: This comment is probably not necessary since the method names are pretty clear

core/Utils.vala
```diff
@@ -720,5 +720,29 @@ namespace Maya.Util {
...
+    
+    public bool show_weeks () {
+        if (GLib.SettingsSchemaSource.get_default ().lookup ("org.pantheon.desktop.wingpanel.indicators.datetime", false) != null) {
```
danrabbit: Since you use this string 4 times it's probably a good idea to just make it a constant

core/Utils.vala
```diff
@@ -720,5 +720,29 @@ namespace Maya.Util {
...
+            GLib.Settings weeks = new GLib.Settings ("org.pantheon.desktop.wingpanel.indicators.datetime");
+            return weeks.get_boolean ("show-weeks");
+        }
+        else {
```
danrabbit: Cuddled else please :)

`} else {`

core/Utils.vala
```diff
@@ -720,5 +720,29 @@ namespace Maya.Util {
...
+            return Settings.SavedState.get_default ().show_weeks;
+        }
+    }
+    
+    public void toggle_show_weeks () {
+        if (GLib.SettingsSchemaSource.get_default ().lookup ("org.pantheon.desktop.wingpanel.indicators.datetime", false)!= null) {
+            GLib.Settings weeks = new GLib.Settings ("org.pantheon.desktop.wingpanel.indicators.datetime");
+            weeks.set_boolean ("show-weeks", !weeks.get_boolean ("show-weeks"));
+        }
+        else {
```
danrabbit: Cuddled else please :)

`} else {`

src/Grid/CalendarView.vala
```diff
@@ -66,7 +67,15 @@ public class Maya.View.CalendarView : Gtk.Grid {
             }
         });
 
-        Settings.SavedState.get_default ().changed["show-weeks"].connect (on_show_weeks_changed);
+        if (GLib.SettingsSchemaSource.get_default ().lookup ("org.pantheon.desktop.wingpanel.indicators.datetime", false) != null) {
+            show_weeks = new GLib.Settings ("org.pantheon.desktop.wingpanel.indicators.datetime");
+            show_weeks.changed["show-weeks"].connect (on_show_weeks_changed);
+            show_weeks.get_value ("show-weeks");
+        }
```
danrabbit: Cuddled else please :)

`} else {`

Thalexander:
In what class would you think it makes sense to be a property? because it needs to be accessed by three classes, and without creating it's own class which was opposed by you initially, I am not sure where it could easily go, unless the whole thing was moved into CalendarView and then the header and WeeksLabels are initialised with a value for showing weeks and then update functions are used to update them, and a new function created in CalendarView for them to call to toggle the setting, but that seems a fairly complex way to handle that...

core/Utils.vala
```diff
@@ -720,5 +720,25 @@ namespace Maya.Util {
     public string get_hexa_color (Gdk.RGBA color) {
         return "#%02X%02X%02X".printf ((uint)(color.red*255), (uint)(color.green*255), (uint)(color.blue*255));
     }
+    
+    public const string show_weeks_schema = "org.pantheon.desktop.wingpanel.indicators.datetime";
```
danrabbit: constants should be uppercase: `SHOW_WEEKS_SCHEMA`

src/Grid/WeekLabels.vala
```diff
@@ -135,7 +136,7 @@ public class WeekLabels : Gtk.Revealer {
         cr.stroke ();
         return false;
     }
-
+
```
danrabbit: looks like a few extra spaces got added here

Thalexander:
That's fine as in do the CalendarView thing or that's fine as in it's fine as it is?

danrabbit:
LGTM! Thank your :tada:

PR #173, opened 2017-10-03
Allow right clicking week labels

Thalexander:
Fixes right clicking week labels to allow hiding them.


Thalexander:
Well it is easier to discover than right clicking the week days.

jeremypw:
But the user would have already discovered that to show the week numbers (which are not shown by default)?

Thalexander:
They're shown by default, org.pantheon.maya.settings.show-weeks has a default of true in the schema.

jeremypw:
Ah, OK, my mistake, they've always been off for me.  Anyway, I have no objection, just noting elementaryos generally discourages too many alternatives ways of doing things.

PR #174, opened 2017-10-06
Update EventButtons directly

### LINKED ISSUE #127: Editing an event causes a duplicate event to show in the grid momentarily
Steps taken:
1. Create an event and save it
2. Double-click the event in the sidebar to edit it
3. After editing, save the event

Observation:
After clicking "Save Changes," the event gets duplicated in the main grid momentarily, then the grid refreshes and the duplicate entry is removed.

Why is this an issue?
This doesn't seem to be a necessary visual feature of the program, and fixing the issue will help polish the editing workflow.

Thalexander:
This replaces the remove and re-add behaviour for updating events and directly updates the event on the button. As a result there is no duplication on the event whilst this is going on. This does add quite a bit of complexity when handling multi-day events.

Fixes #127

### DIFF
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -144,6 +144,17 @@ namespace Maya.Util {
 
         return false;
     }
+    
+    public bool is_day_in_range (DateTime day, Util.DateRange range) {
+        var date = day.get_day_of_year ();
+        
+        foreach (var dt in range) {
+            if (dt.get_day_of_year () == date && dt.get_year () == day.get_year ()) {
+                return true;
+            }
+        }
+        return false;
+    }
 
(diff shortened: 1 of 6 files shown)


Thalexander:
This should probably not be merged until after #171  as this will need to be modified to re-sort the EventButtons after being updated.

tintou:
I'm fine with the current code, the ListBox port might come in another PR if needed

Thalexander:
This correctly now updates the sorting when events update. It would be possible to use the update function when adding events and shrink the code, however I have not done this as it would cause the EventButtons to be added before being sorted so they would appear in one place then get sorted to a new one.

PR #175, opened 2017-10-07
Correct selecting days in other months

### LINKED ISSUE #74: Days of previous / following month only selectable at the very top
Selecting a day of the current month works by just clicking anywhere within the rectangle representing that day. However, days of the previous or following month can only be selected if I click the top part of that rectangle, around the height of the grey number which is the date.

I am running elementary OS 0.3.1 Freya

Launchpad Details: [#LP1523152](https://bugs.launchpad.net/bugs/1523152) Kirue - 2015-12-05 19:29:41 +0000

Thalexander:
Previously GridDays in other months were marked as insensitive, this caused issues when clicking on them and required you to click at the very top (the label) to change month, this resolves this by instead using a css class to style cells that are in other months. A side effect of this push is that the background colour for other months applies, whereas previously the css that should have been applied was not, if this is not a desired result, that rule can easily be removed to just be the grayed out text.

screenshot from 2017-10-08 00 06 03

Fixes #74

### DIFF
--- a/data/style/calendar.css
+++ b/data/style/calendar.css
@@ -18,7 +18,6 @@
 
 @define-color cell_color mix(@bg_color, rgb(255, 255, 255), 0.5);
 @define-color text_color #333;
-@define-color alt_text_color #AEADAE;
 
 /* Header Styles */
 
@@ -38,7 +37,6 @@
 
 .other_month {
     background-color: shade(@cell_color, 0.97);
-    color: @alt_text_color;
 }
 
 .cell:selected {
--- a/data/style/calendar.css
+++ b/data/style/calendar.css
@@ -18,6 +18,7 @@
 
 @define-color cell_color mix(@bg_color, rgb(255, 255, 255), 0.5);
 @define-color text_color #333;
+@define-color alt_text_color #AEADAE;
 
 /* Header Styles */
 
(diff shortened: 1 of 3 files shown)


jeremypw:
Looks good to me but UX team should approve color.

The elementary code style would suggest that a property of GridDay be used instead of a function to set the style i.e:

```
	public bool in_current_month {
		set {
			if (value) {
				get_style_context ().remove_class ("other_month");
			} else {
				get_style_context ().add_class ("other_month");
			}
		}
	}
```
and in Grid update_day ():
```
        day.in_current_month = (new_date.get_month () == month_start.get_month ());
```

danrabbit:
Ah yeah that's no good. There's actually a Gtk style class constant for this [`Gtk.STYLE_CLASS_DIM_LABEL`](https://valadoc.org/gtk+-3.0/Gtk.STYLE_CLASS_DIM_LABEL.html)

Thalexander:
I'm not as fond of that class, the text feels too bold still. Not sure if I am using it wrong though.
screenshot from 2017-10-17 22 22 02

danrabbit:
You can file a new issue in the stylesheet if you'd like :p But we should definitely use this style class to remain consistent throughout the system

PR #178, opened 2017-10-14
Allow right clicking EventButtons

Thalexander:
This adds the ability to right click events on the grid to edit or remove them, this matches the behaviour of right clicking events in the agenda view.

### DIFF
--- a/src/Grid/EventButton.vala
+++ b/src/Grid/EventButton.vala
@@ -53,10 +53,17 @@ public class Maya.View.EventButton : Gtk.Revealer {
                 edit_event ();
             } else if (event.type == Gdk.EventType.BUTTON_PRESS && event.button == Gdk.BUTTON_SECONDARY) {
                 if (menu == null) {
+                    E.Source src = comp.get_data ("source");
+
                     menu = new Gtk.Menu ();
                     menu.attach_to_widget (this, null);
                     var edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
                     var remove_item = new Gtk.MenuItem.with_label (_("Remove"));
+                    if (src.writable != true && Model.CalendarModel.get_default ().calclient_is_readonly (src) != false) {
+                        edit_item.sensitive = false;
+                        remove_item.sensitive = false;
+                    }
+
                     edit_item.activate.connect (() => { edit_event (); });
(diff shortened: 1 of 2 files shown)


danrabbit:
If this is the same menu, maybe the menu should be its own class so we're not duplicating the code and we keep its items in sync

src/Grid/EventButton.vala
```diff
@@ -49,14 +50,31 @@ public class Maya.View.EventButton : Gtk.Revealer {
         event_box.add (internal_grid);
         event_box.button_press_event.connect ((event) => {
             if (event.type == Gdk.EventType.2BUTTON_PRESS && event.button == Gdk.BUTTON_PRIMARY) {
-                E.Source src = comp.get_data ("source");
-                if (src.writable == true && Model.CalendarModel.get_default ().calclient_is_readonly (src) == false) {
-                    edition_request ();
-                    return true;
+                edit_event ();
+            } else if (event.type == Gdk.EventType.BUTTON_PRESS && event.button == Gdk.BUTTON_SECONDARY) {
+                if (menu == null) {
```
jeremypw: It is unnecessary to test for null in creation method (and may not give intended result).  Menu should always be created as it is an instance variable.

src/Grid/EventButton.vala
```diff
@@ -49,14 +50,31 @@ public class Maya.View.EventButton : Gtk.Revealer {
...
+                    E.Source src = comp.get_data ("source");
+
+                    menu = new Gtk.Menu ();
+                    menu.attach_to_widget (this, null);
+                    var edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
+                    var remove_item = new Gtk.MenuItem.with_label (_("Remove"));
+                    if (src.writable != true && Model.CalendarModel.get_default ().calclient_is_readonly (src) != false) {
```
jeremypw: This test for writability is repeated in edit_event and remove event.  If it could change during the lifetime of the instance, it could be made a property e.g. 
```
private bool can_edit {
    get {
         ...
    }
}
```
Otherwise it could just be determined once on creation and assigned to a variable.

src/Grid/EventButton.vala
Thalexander: I was following the existing code style, I have now changed this.

PR #179, opened 2017-10-14
Remove email notifications

### LINKED ISSUE #177: Email notifications are not implemented
Email notifications can be added in the event editor but the daemon does not implement them. I think there would need to be some new UI to handle email settings to implement this.

Thalexander:
This removes email notifications as an option as it isn't implmeneted. It leaves the ComboBox to select notification type for the potential of re-adding it, when it is implemented. 

Closes #177 (Although we probably want to create an issue to add email notifications)

### DIFF
--- a/daemon/Daemon.vala
+++ b/daemon/Daemon.vala
@@ -147,13 +147,6 @@ namespace MayaDaemon {
                         }
                     }
                     continue;
-                case (E.CalComponentAlarmAction.EMAIL):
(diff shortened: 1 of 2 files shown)


danrabbit:
Probably should remove the combobox as well since there's only one option

Thalexander:
I have removed the Combobox and replaced it with a label, but tried to keep the overall appearance the same.
screenshot from 2017-10-15 02 05 51

PR #191, opened 2017-11-04
Add remove occurrence button

Thalexander:
When an event has repetitions this makes it so right clicking on them in the grid includes remove series and remove occurrence buttons rather than the generic remove, this is to create a simpler way to add exception dates.

### DIFF
--- a/src/Grid/EventButton.vala
+++ b/src/Grid/EventButton.vala
@@ -66,9 +66,9 @@ public class Maya.View.EventButton : Gtk.Revealer {
 
                 Gtk.MenuItem remove_item;
                 if (comp.has_recurrences ()) {
-                    remove_item = new Gtk.MenuItem.with_label (_("Remove series"));
-                    Gtk.MenuItem exception_item = new Gtk.MenuItem.with_label (_("Remove occurrence"));
-                    exception_item.activate.connect (() => { add_exception (); });
+                    remove_item = new Gtk.MenuItem.with_label (_("Remove Series"));
+                    Gtk.MenuItem exception_item = new Gtk.MenuItem.with_label (_("Remove Occurrence"));
+                    exception_item.activate.connect (add_exception);
                     exception_item.sensitive = sensitive;
                     menu.append (exception_item);
                 } else {
@@ -76,7 +76,7 @@ public class Maya.View.EventButton : Gtk.Revealer {
                 }
 
                 remove_item.sensitive = sensitive;
-                remove_item.activate.connect (() => { remove_event (); });
+                remove_item.activate.connect (remove_event);
                 menu.append (remove_item);
 
                 menu.popup (null, null, null, event.button, event.time);
--- a/src/Grid/EventButton.vala
+++ b/src/Grid/EventButton.vala
@@ -25,12 +25,14 @@
 public class Maya.View.EventButton : Gtk.Revealer {
     public signal void edition_request ();
     public E.CalComponent comp {get; private set;}
+    private GLib.DateTime date;
     private Gtk.EventBox event_box;
     private Gtk.Grid internal_grid;
     Gtk.Label label;
 
-    public EventButton (E.CalComponent comp) {
+    public EventButton (E.CalComponent comp, GLib.DateTime date) {
         this.comp = comp;
+        this.date = date;
         transition_type = Gtk.RevealerTransitionType.CROSSFADE;
         internal_grid = new Gtk.Grid ();
         internal_grid.column_spacing = 6;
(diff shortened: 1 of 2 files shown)


src/Grid/EventButton.vala
```diff
@@ -55,17 +57,26 @@ public class Maya.View.EventButton : Gtk.Revealer {
...
+                Gtk.MenuItem edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
+                edit_item.activate.connect (() => { edition_request (); });
+                edit_item.sensitive = sensitive;
+                menu.append (edit_item);
+
+                Gtk.MenuItem remove_item;
+                if (comp.has_recurrences ()) {
+                    remove_item = new Gtk.MenuItem.with_label (_("Remove series"));
+                    Gtk.MenuItem exception_item = new Gtk.MenuItem.with_label (_("Remove occurrence"));
+                    exception_item.activate.connect (() => { add_exception (); });
```
danrabbit: This could be written more simply as `exception_item.activate.connect (add_exception);`

src/Grid/EventButton.vala
```diff
@@ -55,17 +57,26 @@ public class Maya.View.EventButton : Gtk.Revealer {
...
-                    edit_item.sensitive = false;
-                    remove_item.sensitive = false;
+                Gtk.MenuItem edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
+                edit_item.activate.connect (() => { edition_request (); });
+                edit_item.sensitive = sensitive;
+                menu.append (edit_item);
+
+                Gtk.MenuItem remove_item;
+                if (comp.has_recurrences ()) {
+                    remove_item = new Gtk.MenuItem.with_label (_("Remove series"));
```
danrabbit: I'm not sure these are super clear labels. Maybe it would be more clear to have "Remove This Event" and "Remove All Future Events"?

Either way, menu items should be Title Case like buttons :)

src/Grid/EventButton.vala
```diff
@@ -55,17 +57,26 @@ public class Maya.View.EventButton : Gtk.Revealer {
...
 
-                var edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
-                var remove_item = new Gtk.MenuItem.with_label (_("Remove"));
+                bool sensitive = src.writable == true && Model.CalendarModel.get_default ().calclient_is_readonly (src) == false;
 
-                if (src.writable != true && Model.CalendarModel.get_default ().calclient_is_readonly (src) != false) {
-                    edit_item.sensitive = false;
-                    remove_item.sensitive = false;
+                Gtk.MenuItem edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
+                edit_item.activate.connect (() => { edition_request (); });
```
danrabbit: This could be `edit_item.activate.connect (edition_request);`

Thalexander: edition_request is a signal, and Vala does not seem to like connecting it in this way, changing the others to this form was fine though.

src/Grid/EventButton.vala
Thalexander: The problem is we refer to events throughout the app as containing all instances, so we have "Remove Event" throughout removing the entire series, I could accept changing "Remove Series" to "Remove Event" but that seems a bit confusing... "Remove All Future Events" I am very opposed to as that would in my opinion just remove future events, which is a very different behaviour to what this is doing.

danrabbit: Can you clarify which each of the buttons do? To me, the terms “series” and “occurrence” aren’t clear. I don’t know what those mean or what will happen when I select one of these items

Thalexander: Events can repeat so for example an event can repeat every tuesday and thursday, but they still are the same event just on multiple dates (or datetimes), currently when you remove an event we delete all of the dates, in this pull I am using Series to refer to the whole set of dates referred to by an event which is defined by the recurrence rule (and some other properties). Whereas the Occurence term I am using to refer to on specific date that an event occurs on, so just one of the Tuesdays because you can't do the meeting that day (in reality the meeting would probably be moved which is why we should support RDATES as well as EXDATES) so an exception is added to say that the event will not occur on that date.

Thalexander: For a comparison Timepage has "Delete this event" in place of the "Remove Occurrence" and "Delete all occurrences" in place of "Remove Series" however we may need to redesign existing parts of the interface to treat an event as being just one date of an event... iOS seems to similarly treat each date as an event, but then only let you modify future occurrences (and this behaviour seems quite unpredictable even with standard complying events.) Blackberry uses Series and Occurrence like this. Gnome Calendar (at least the version in Loki) only seems to delete Occurrences individually as "Delete Event", but then it also doesn't seem to properly support repeats and they have to be created externally...

danrabbit: If we have a precedent in another part of the UI, we should probably stick to using the same labels so we stay consistent with ourselves

Thalexander: Another related thing is we may want to stick to either 'remove' or 'delete' if trying to be consistent, but that is another problem.

PR #206, opened 2017-12-04
Delete AUTHORS file

Thalexander:
Outdated, GitHub provides more up to date contributor information.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,0 @@
-Copyright (c) 2011-2012 by Maya Developers
-
-Written by Maya Developers:
-* Maxwell Barvian <maxwell@elementaryos.org>
-* Jaap Broekhuizen <jaapz.b@gmail.com>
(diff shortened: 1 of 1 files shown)


cassidyjames:
@tintou there should probably be an effort at discussion before things like this are approved and merged. For example, our official developer guide still points people to create and `AUTHORS` file, and most of our other projects still include them. Have we decided to update the docs and all of our repos to match? Is the AUTHORS file useful outside of the GitHub web interface? Do downstream projects use the AUTHORS file to credit contributors?

cassidyjames:
Actually, it could be useful to automatically generate the AUTHORS file from a set of rules from the commit history as well. That way it's 

1. Up to date
2. Keeps the strong convention from FOSS
3. Never has to be worried about again
4. Available locally/offline

tintou:
@cassidyjames It's an old convention, most projects just redirect to git log. There is no need to even have this file, it's only a requirement when using Autotools.

PR #209, opened 2017-12-23
Use the next whole hour for the default time

### LINKED ISSUE #208: The default time for new events should be a full hour
When I add a new event, the pre-entered time is set to the current time, which makes absolutely no sense as most appointments are most likely to start at XX:00 (or XX:15, XX:30, XX:45). So we could spare the user from the extra step of changing the minutes back to :00 by setting the default time to the next full hour or something like that.
Maybe a quick selection menu of :00, :15, :30, :45 like there is in Thunderbird Lightning or other calendar apps would be useful, too.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/52979530-the-default-time-for-new-events-should-be-a-full-hour?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Thalexander:
Sets the default time for new events to the next whole hour, however does not increment the date as that would create unexpected results for the user when they select a day and get the next day.

Fixes #208


Thalexander:
We need to keep the event on the same day as when they click a day it should always open that day, so the alternative to wrapping back to 0 is to cap it so that it will round down to 23. Either way the from needs to go into the past as far as I can tell, the issue with 23 is that the to ends up as the next day, which if they select a day may require them changing the date of the to, or more likely not noticing.

jeremypw:
I would prefer that the hour is clamped to a max of 23 rather than go into the past. I agree that the day should not change. However, maybe design team input is required.

jeremypw:
Any conclusion from @elementary/ux ?:
>I  would prefer that the hour is clamped to a max of 23 rather than go into the past. I agree that the day should not change. However, maybe design team input is required.

cassidyjames:
I agree that it should default to the day that was clicked, and I think clamping to 23:00 is sane. :+1:

CodingSpiderFox:
This PR is quite old. Any updates on this?

jeremypw:
> This PR is quite old. Any updates on this?

Still waiting for sign off by @elementary/ux

jeremypw:
@cassidyjames So the start-time is 23.00 and the end-time is 00.00 the following morning?

cassidyjames:
@jeremypw yeah I think so

src/EventEdition/InfoPanel.vala
```diff
@@ -273,10 +273,29 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
...
-            from_time_picker.time = new DateTime.now_local ();
-            to_date_picker.date = parent_dialog.date_time;
-            to_time_picker.time = new DateTime.now_local ().add_hours (1);
+
+            /* Default event duration of one hour, changing day if required */
+            if (now_before_2300) {
+                to_time_picker.time = time.add_hours (1);
+                to_date_picker.date = parent_dialog.date_time;
+            } else { /* Default end time is 00.00 (12.00 AM) next morning*/
+                time = time.add_hours (1);
```
jlnr: Mostly curious, why is this on its own line? It seems like it could be merged with the assignment below as in the other if branches.

jeremypw: Its just an oversight. I'll fix it.

PR #215, opened 2018-01-19
Style New Calendar as a menuitem

danrabbit:
Style "New Calendar..." in the app menu as a menuitem.

While we're here, do a little organization/cleanup


Thalexander:
Before:
screenshot from 2018-01-19 22 54 08

After:
screenshot from 2018-01-19 22 53 09

Does this rely on changes not in Loki, because it doesn't seem to be an improvement on my system?

jeremypw:
You can emulate the original appearance by removing the MENUITEM style class, adding a margin to the button and removing the horizontal separator.  However, I am not sure what the intended appearance is.

danrabbit:
screenshot from 2018-02-09 09 21 49

Should look like this. You probably need to be on Juno

PR #218, opened 2018-02-11
Focus GridDay on clicking EventButton

### LINKED ISSUE #213: Clicking once on event should activate day?
When clicking on an event, instead of the top part or empty part of a day, nothing happens. Shouldn't a single click on an event also navigate to the day?

Thalexander:
Fixes #213


jeremypw:
You do not need to create a new signal - just return false from the event button press handler to pass the event up the chain.

Thalexander:
I thought there had to be something like that, thanks.

PR #249, opened 2018-07-03
Full translation of French and Dutch.

Ordissimo:


DDieter:
The way we do translations is by using [Weblate](https://l10n.elementary.io/). I am responsible for the Dutch translations so I suggest you add your translations on Weblate (you can easily login with your Github account) and then I will check them.

You can do the same for the French translations.

Ordissimo:
I just watched Welate.
We did the translation with a professional translator.
Weblate, seems to me very well. Say it from the reception.
Unfortunately the project did not update these po. Go find the chains that were not, the one that has been changed will take me too much time.
Sorry.

PR #251, opened 2018-07-04
Port to meson build system

tintou:
Because of future and stuff


codygarver:
Found them, they're named `elementary-calendar.vapi` and `elementary-calendar.h` now instead of `io.elementary.calendar.*`

tintou:
@codygarver Yeah, the issue here is that the pkg-config file was "libmaya-calendar", the .po file was libelementary-calendar and the headers were RDNN so let's stuck with libelementary-calendar everywhere.

worldofpeace:
Doesn't this need a post_install script like [this](https://github.com/elementary/code/blob/master/meson/post_install.py)?

worldofpeace:
I added it here #254

PR #252, opened 2018-07-07
Drop gconf-2.0

zaitor:
Drop long obsolete gconf-2.0 from core packages


tintou:
Yeah, no idea why it got there but it's fixed in #251

danrabbit:
Closing because it was superseded by #251

PR #276, opened 2018-10-10
Remove cairo draws

dahenson:
This branch primarily removes Cairo draw functions that were used to draw the borders of the grid. Instead, there is CSS to take care of the drawing.

The only funkiness is in the `CalendarView.vala` where the top left corner had to be filled with a blank label widget in order to draw a border next to the Sunday label. If the Sunday label draws the border, it is not in line with the week number border. Additionally, the blank widget needs to be shown and hidden when the week numbers are revealed/hidden, so it is slightly more complicated than I wish it could be.

What's important here is that it works and it uses less code and more CSS to draw the borders. If anyone has questions, feel free to hit me up.

### DIFF
--- a/data/style/calendar.css
+++ b/data/style/calendar.css
@@ -42,6 +42,10 @@
     border-left: solid 1px alpha(#000, 0.25);
 }
 
+.cell.firstcol {
+	border-left: none;
+}
+
 .other_month {
     background-color: shade(@cell_color, 0.97);
 }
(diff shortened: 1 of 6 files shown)


data/style/calendar.css
```diff
@@ -28,11 +28,26 @@
 			from(@cell_color), to(mix(@bg_color, rgb(255, 255, 255), 0.8)));
 }
 
+.daylabel {
+    padding: 3px;
+    border-left: solid 1px alpha(#000, 0.25);
+}
+
+.daylabel.firstcol {
```
danrabbit: Hm, is there no way to use :first-child for this?

PR #281, opened 2018-10-19
Change accent color to lime

cassidyjames:
Pretty subtle, but carries the green calendar brand into the app more, which I think I like.

screenshot from 2018-10-19 16 29 44
screenshot from 2018-10-19 16 30 19


lainsce:
Don't forget current day selection, it's still blue :P

cassidyjames:
Using the accent variable gives me this, which I wasn't as hot on

image

lainsce:
Looks nice to me, to be honest

danrabbit:
Hm something is awry here. I'm not seeing the accent color on switches etc, only for the selected day

cassidyjames:
@danrabbit ah yeah interesting. The same definition works in the inspector but not here. I'm guessing it's however the CSS is being loaded?

cassidyjames:
@danrabbit since this is an app-wide change and not just for the calendar view, I added a new Application stylesheet and loaded it for the app.

cassidyjames:
screenshot from 2018-12-12 16 42 31
screenshot from 2018-12-12 16 42 37
screenshot from 2018-12-12 16 43 00

PR #285, opened 2018-11-02
SourceDialog: misc cleanup

danrabbit:
* Adjust margins
* Remove extra widget
* Organize
* Remove unnecessary casts
* Bump copyright


danrabbit:
Yeah just because they’re redundant and obvious from the variable names

PR #286, opened 2018-11-02
Redesign removable source items

danrabbit:
* Don't noshow deleted message since that causes popover resize jank
* Update the deleted message label when source item names change
* Don't hide buttons when items are not actionable. Instead make insensitive and give explanatory tooltips.
* Give the delete action a destruction action class
* Move close button to the left in undo message
* Use a over/under left-swipe kind of animation
* Remove unnecessary casts
* Organize


danrabbit:
@dahenson Resolved by creating a label, xaligning the label, and then packing that label into the button

PR #292, opened 2018-11-13
SourceItem: style checkboxes with calendar colors

danrabbit:
## BEFORE

screenshot from 2018-11-13 13 54 56 2x

## AFTER

screenshot from 2018-11-13 13 48 47 2x

### DIFF
--- a/src/SourceDialog/SourceItem.vala
+++ b/src/SourceDialog/SourceItem.vala
@@ -52,9 +52,6 @@ public class Maya.View.SourceItem : Gtk.ListBoxRow {
         label = source.dup_display_name ();
         location = Maya.Util.get_source_location (source);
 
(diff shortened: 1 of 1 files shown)


peteruithoven:
Elegant indeed. But shouldn't the white arrow then also adapt it's color based on it's background? I thought we had some utils somewhere to help with that, making sure there is enough contrast.

cassidyjames:
Ah yeah on one computer here the built-in event colors are super light. Where are these being set?

screenshot from 2018-11-13 15 18 32

cassidyjames:
@peteruithoven we don't have a util afaik, I reported that at https://github.com/elementary/granite/issues/244

peteruithoven:
Ah yes! Sounds like we have another reason to include that in Granite.

PR #297, opened 2018-12-06
EventButton: misc clean up

danrabbit:
* Bump copyright
* Organize
* Reduce variable scope
* Remove eventbox that seems to do nothing. Maybe @tintou has information otherwise?
* GObject style
* Implicit typing


jeremypw:
Not sure, but I think the fake label event box is so that a different scroll event handler can be triggered if scrolling on the fake label area as opposed to the real label area (by default, events go the eventbox's children first).

tintou:
Yeah the eventbox was probably there because it was the only widget that received all the signals and could be drawn at the same time.

PR #298, opened 2018-12-06
ImportDialog: subclass Granite.MessageDialog

danrabbit:
This dialog is kind of wonky if you want to test it. You need to have calendar already running and then you can click an .ics file in FIles to see the dialog in Calendar

## BEFORE

screenshot from 2018-12-06 14 25 56 2x

## AFTER

screenshot from 2018-12-06 14 24 11 2x


cassidyjames:
Could probably simplify a bit?

> ### Select a Calendar to Import Into
>Events from "sample.ics" will be merged with this calendar:

And I think for AP Style we'd keep "a" and "to" lowercase?

PR #301, opened 2018-12-10
GridDay.vala: misc cleanup

danrabbit:
* GObject-style
* Organize
* Make date property public and connect to notify instead of setter
* Reduce variable scope
* Set margins in the normal Gtk way
* Bump copyright header

### DIFF
--- a/src/Grid/Grid.vala
+++ b/src/Grid/Grid.vala
@@ -167,7 +167,7 @@ public class Grid : Gtk.Grid {
 
         day.in_current_month = new_date.get_month () == month_start.get_month ();
 
-        day.update_date (new_date);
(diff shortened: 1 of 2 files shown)


cassidyjames:
@danrabbit it seems that this util is unused now, so I'd say remove it in this PR?

https://github.com/elementary/calendar/blob/3286f7af9b38532ac36ca391e484126530fccf43/core/Utils.vala#L561-L567

danrabbit:
@cassidyjames I would, but it'll break CI because of the way packaging is done. So my plan is to merge this, merge a packaging branch that removes the check for this method, then merge a branch that removes it

PR #311, opened 2018-12-11
Put color radios in their own grid

### LINKED ISSUE #310: Grid alignment off with colored radio buttons
When you choose a type other than local, all the extra widgets are in the same column as the first radio button, causing this:

screenshot from 2018-12-11 11 18 08

When I want a group of widgets to be treated like a a single widget, I usually attach them all to their own grid and then attach that grid to the layout grid to prevent stuff like this.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67358267-grid-alignment-off-with-colored-radio-buttons?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Since for layout purposes we want to treat the collection of radios like a single widget, this adds them all to their own grid and simplifies some of the other grid layout.

screenshot from 2018-12-11 11 31 03

Fixes #310

### DIFF
--- a/src/SourceDialog/SourceDialog.vala
+++ b/src/SourceDialog/SourceDialog.vala
@@ -6,12 +6,12 @@
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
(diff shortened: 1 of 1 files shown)


src/SourceDialog/SourceDialog.vala
```diff
@@ -169,6 +169,18 @@ public class Maya.View.SourceDialog : Gtk.Grid {
...
+        color_grid.add (color_button_red);
+        color_grid.add (color_button_orange);
+        color_grid.add (color_button_yellow);
+        color_grid.add (color_button_green);
+        color_grid.add (color_button_blue);
+        color_grid.add (color_button_purple);
+        color_grid.add (color_button_brown);
+        color_grid.add (color_button_slate);
+
+
```
danrabbit:
```suggestion
```

alainm23:
You can also add a color in hexadecimal as in Planner.
color

danrabbit:
@alainm23 wanna file an issue report for that?

PR #315, opened 2018-12-11
AgendaView: misc cleanup

danrabbit:
* Bump copyright header
* Row and column spacing are 0 by default, don't need to specify
* Use `xalign` property instead of `set_alignment ()`
* Use Granite style class constants
* Remove unnecessary sep `hexpand`s
* `add` instead of `attach`
* Explicit method scope
* Remove unused method

### DIFF
--- a/src/AgendaView.vala
+++ b/src/AgendaView.vala
@@ -1,6 +1,5 @@
-// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/*-
- * Copyright (c) 2011-2015 Maya Developers (http://launchpad.net/maya)
+/*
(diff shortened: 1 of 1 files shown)


src/AgendaView.vala
```diff
@@ -1,6 +1,5 @@
-// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/*-
- * Copyright (c) 2011-2015 Maya Developers (http://launchpad.net/maya)
+/*
+ * Copyright (c) 2011-2018 elementary, Inc. (https://elementary.io)
```
tintou:
```suggestion
 * Copyright 2011-2018 elementary, Inc. (https://elementary.io)
```

PR #316, opened 2018-12-11
CalendarButton: subclass MenuButton, cleanup

danrabbit:
* Bump copyright header
* Subclass Gtk.MenuButton instead of ToggleButton
* Use built-in popover and get rid of code about showing and hiding it
* Use new features of Gtk for scrollbox max content height
* Make a single-use function into a lambda
* Reduce scope of `list_box`

### DIFF
--- a/src/Widgets/CalendarButton.vala
+++ b/src/Widgets/CalendarButton.vala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2018 elementary, Inc. (https://elementary.io)
+ * Copyright 2014-2018 elementary, Inc. (https://elementary.io)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


src/Widgets/CalendarButton.vala
```diff
@@ -61,46 +57,25 @@ public class Maya.View.Widgets.CalendarButton : Gtk.ToggleButton {
...
+        scrolled.max_content_height = 300;
+        scrolled.propagate_natural_height = true;
+        scrolled.show_all ();
 
         popover = new Gtk.Popover (this);
         popover.width_request = 310;
         popover.add (scrolled);
 
-        list_box.add.connect ((widget) => {
+        list_box.add.connect (() => {
```
tintou: we can remove this and do `row.show_all ()` at row creation

src/Widgets/CalendarButton.vala
```diff
@@ -118,25 +93,22 @@ public class Maya.View.Widgets.CalendarButton : Gtk.ToggleButton {
 
         list_box.row_activated.connect ((row) => {
             current_source = ((CalendarGrid)row.get_child ()).source;
+            popover.popdown ();
```
tintou: unsure that you really want this

danrabbit: Yeah to really mimic the expected combobox behavior we want the popover to hide after we make a selection

PR #322, opened 2018-12-12
Try to assign icons to events

danrabbit:
Looks for keywords, localized and in English and tries to assign an icon

screenshot from 2018-12-12 10 50 50 2x

### DIFF
--- a/src/Widgets/AgendaEventRow.vala
+++ b/src/Widgets/AgendaEventRow.vala
@@ -37,6 +37,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
     public bool is_multiday { public get; private set; default = false; }
     public Gtk.Revealer revealer { public get; private set; }
 
+    private Gtk.Image event_image;
     private Gtk.Label name_label;
     private Gtk.Label datatime_label;
     private Gtk.Label location_label;
@@ -59,7 +60,8 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
 
         E.SourceCalendar cal = (E.SourceCalendar)source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
 
-        var event_image = new Gtk.Image.from_icon_name ("office-calendar-symbolic", Gtk.IconSize.MENU);
+        event_image = new Gtk.Image.from_icon_name ("office-calendar-symbolic", Gtk.IconSize.MENU);
+        event_image.pixel_size = 16;
         event_image.valign = Gtk.Align.START;
 
         name_label = new Gtk.Label ("");
@@ -159,6 +161,14 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
         unowned iCal.Component ical_event = event.get_icalcomponent ();
         summary = ical_event.get_summary ();
         name_label.set_markup (Markup.escape_text (summary));
+
+        var event_name = name_label.label.down ();
+
+        if (_("birthday") in event_name || "birthday" in event_name) {
+            event_image.icon_name = "event-birthday-symbolic";
+        } else if (_("flight") in event_name || "flight" in event_name) {
+            event_image.icon_name = "airplane-mode-symbolic";
+        }
 
         DateTime start_date, end_date;
         Util.get_local_datetimes_from_icalcomponent (ical_event, out start_date, out end_date);


cassidyjames:
This feels _a little_ dirty, but at the same time, I'm 99% sure it's the same "magic" Google Calendar is using to add pretty illustrations to events automatically. The nice thing is that they do stuff like inferring a movie viewing (probably based in part on the location, but possibly also other language), hotel stay, meetings, etc. and add relevant illustrations to events. If it ends up being relatively reliable, I think it would be a really nice touch.

cassidyjames:
@danrabbit if we're going to pursue this in any meaningful way, I wonder if it would make sense to pull the phrases out into their own array somehow, so it's a little less messy all inline.

src/Widgets/AgendaEventRow.vala
```diff
@@ -161,6 +163,28 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
         summary = ical_event.get_summary ();
         name_label.set_markup (Markup.escape_text (summary));
 
+        var event_name = name_label.label.down ();
+
+        if (_("dinner") in event_name || _("reservation") in event_name || _("steakhouse") in event_name) {
```
cassidyjames:
```suggestion
        if (_("breakfast") in event_name || _("brunch") in event_name || _("lunch") in event_name || _("supper") in event_name || _("dinner") in event_name || _("reservation") in event_name || _("steakhouse") in event_name) {
```

src/Widgets/AgendaEventRow.vala
```diff
@@ -161,6 +163,28 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+            event_image.icon_name = "event-food-symbolic";
+        } else if (_("birthday") in event_name) {
+            event_image.icon_name = "event-birthday-symbolic";
+        } else if (_("flight") in event_name) {
+            event_image.icon_name = "event-flight-symbolic";
+        } else if (_("wedding") in event_name) {
+            event_image.icon_name = "event-wedding-symbolic";
+        } else if (_("tax") in event_name || _("jury") in event_name || _("court") in event_name) {
+            event_image.icon_name = "event-legal-symbolic";
+        } else if (_("appointment") in event_name) {
```
cassidyjames:
```suggestion
        } else if (_("appointment") in event_name || _("meeting") in event_name) {
```

src/Widgets/AgendaEventRow.vala
```diff
@@ -161,6 +163,28 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+            event_image.icon_name = "event-appointment-symbolic";
+        } else if (_("road trip") in event_name) {
```
cassidyjames:
```suggestion
        } else if (_("drive") in event_name || _("road trip") in event_name) {
```

src/Widgets/AgendaEventRow.vala
```diff
@@ -161,6 +163,28 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+            event_image.icon_name = "event-driving-symbolic";
+        } else if (_("call") in event_name) {
```
cassidyjames:
```suggestion
        } else if (_("call") in event_name || _("phone") in event_name) {
```

danrabbit:
@cassidyjames Yeah I think probably what would be better is some kind of ranking system so that we can say "Oh we got 2 keyword hits for birthday but we got 3 keyword hits for dinner, so this is more likely to be a food event". So far I was just trying to get together some icons and prove the concept.

cassidyjames:
@danrabbit it looks like GeoClue can be yet-another-hint but it lacks a _lot_ of specific location types, like a movie theater from your example. So I think keyword matching might still be useful there.

cassidyjames:
@danrabbit another few I've noticed Google picking up, even without any location data:

- "drinks" or "cocktails"
- "dentist" and related keywords

src/Widgets/AgendaEventRow.vala
```diff
@@ -155,6 +171,160 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+            _("bar"),
+            _("cocktails"),
+            _("drinks"),
+            _("happy hour"),
+        };
+
+        string[] driving_keywords = {
+            _("car"),
+            _("drive"),
+            _("rental"),
```
cassidyjames: I'd be worried this could trigger for things like an Airbnb rental

jeremypw:
You should consider i18n issues - some English keywords are ambiguous when in isolation so some translator hints would be needed.

cassidyjames:
@danrabbit another one I noticed in Google Calendar recently was for "showers," i.e. "baby shower" and "wedding shower". Could probably just use a gift in those instances?

snowparrot:
I dont think fixed hard-coded arrays are a good solution regarding the translation issue.

In some languages there are expression you cant translate into English. For this
you cant find on that hard coded list. (like e.g. German has the expression "Essen gehen" which means "going out for food" but nobody says it in English). On a German list it should be included but not an English list.

I would propose a file where you save the list. In opposite to .po-files it should be possible more phrases for each category than the English list has.

danrabbit:
@snowparrot would a semicolon delimited list be acceptable? Like the keywords in .desktop files? So:

`_("Keyword;Key;Word;Key Phrase;")`

snowparrot:
If it solves a problem it is a solution :P 
I guess it is very important to document it.

jeremypw:
Probably best to leave location info for another PR and concentrate on getting keywords right (with possibility of additional location info in mind).  I would use extendable ArrayLists for keywords as this would allow the possibility of the user adding there own to the built in ones.  This is probably the only practical way to deal with language specific phrases.  Would need a dialog to enter them (although they could be stored in plain text files).

cassidyjames:
@jeremypw since this is not core or critical functionality _at all_ and is just some extra user delight, I don't think we want to make it user-configurable. That sounds like a lot of shit work for extremely little benefit, and opens the possibility of people adding phrases without fully understanding how the matching works, winding up with weird and wrong matches.

jeremypw:
@cassidyjames Fair enough.  I agree we do not need to sweat this too much - it is basically cosmetic.

PR #325, opened 2018-12-12
Replace SourceItemHeader with Granite.HeaderLabel

danrabbit:
Fixes a bug where these header items can be selected


danrabbit:
@cassidyjames this is in the calendar chooser in the new event dialog

PR #326, opened 2018-12-12
DynamicSpinner: misc cleanup

danrabbit:
* Reduce variable scope
* Whitespace
* Use MenuButton
* Center button vertically 
* Make the stop icon symbolic


jeremypw:
Any idea how to get this spinner to show itself?   It seems odd that it is continually spinning but mostly hidden.

danrabbit:
@jeremypw it'll show if you're loading in remote calendars like a public icloud calendar or something

PR #328, opened 2018-12-12
Replace make_label with Granite.HeaderLabel

danrabbit:


cassidyjames:
The alignment feels a little weird now; before, the titles were closer to their respective widgets.

### master

image

### no-make-label

image

cassidyjames:
I guess I like the increased breathing room above the titles, but I feel like they look weird being farther away from their respective widgets.

cassidyjames:
@danrabbit setting the row spacing to 0 seems to better match the old design. Thoughts?

screenshot from 2018-12-13 12 38 35

PR #331, opened 2018-12-12
Mix instead of alpha/shade for event colors

cassidyjames:
Keeps largely the same look, but mixes with theme variables instead of just shading or using alpha. The advantage is that it's more flexible for light-on-dark versus only working for dark-on-light.

### DIFF
--- a/data/style/AgendaEventRow.css
+++ b/data/style/AgendaEventRow.css
@@ -1,7 +1,7 @@
 .event {
-    background: alpha(@accent_color, 0.15);
+    background: mix(@accent_color, @base_color, 0.85);
     border-radius: 3px;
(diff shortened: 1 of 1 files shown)


danrabbit:
Hrm well it's definitely way less saturated doing a mix which doesn't look as nice.

TBH I don't like this trying to make the dark style work thing. That's just not how the app was designed. I'd rather have nicer colors

cassidyjames:
@danrabbit ah, yeah because the foreground color is not pure black or white. That makes sense. I'm mostly looking at this as a proof of concept (and trying to keep it from being totally broken in a dark style), but I agree if we were to actually support a light and dark style in Calendar we'd want to tailor the styles for each.

PR #337, opened 2018-12-13
Daemon: Use GLib.Notification

### LINKED ISSUE #17: Replace libnotify by GLib.Notification
GLib can now do it natively so no need for an external library (and conditional compilation)

Launchpad Details: [#LP1530631](https://bugs.launchpad.net/bugs/1530631) Corentin Noël - 2016-01-03 10:33:57 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44229089-replace-libnotify-by-glib-notification?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #17


daemon/Daemon.vala
```diff
@@ -200,28 +191,16 @@ namespace MayaDaemon {
...
 
-        notification.set_urgency (Notify.Urgency.NORMAL);
-
-        try {
-            notification.show ();
-        } catch (GLib.Error err) {
-            warning ("Could not show notification: %s", err.message);
-        }
-#endif
+        GLib.Application.get_default ().send_notification (uid, notification);
```
jeremypw: This does not work because Daemon.vala is not a GLib.Application (at the moment).

jeremypw:
Ignore comment about libexec.  Binaries here are not expected to be in the user's path (so I had trouble launching it from the terminal for testing).  The daemon should be installed here.

PR #338, opened 2018-12-13
Make calendar color rounded

cassidyjames:
Instead of the awkward rectangle

### Before:

screenshot from 2018-12-13 12 53 58 2x

### After:

image

### DIFF
--- a/core/Utils.vala
+++ b/core/Utils.vala
@@ -639,8 +639,7 @@ namespace Maya.Util {
         string style = """
             .cal_color {
                 %s: %s;
-                border-radius: 999px;
-                padding: 0 0.75em;
+                border-radius: 50%;
             }
         """;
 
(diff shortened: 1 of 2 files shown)


core/Utils.vala
```diff
@@ -636,7 +636,13 @@ namespace Maya.Util {
     }
 
     public void style_calendar_color (Gtk.Widget widget, string color, bool background = false) {
-        string style = ".cal_color { %s: %s }";
+        string style = """
+            .cal_color {
+                %s: %s;
+                border-radius: 999px;
```
danrabbit: why not 50%?
```suggestion
                border-radius: 50%;
```

core/Utils.vala
```diff
@@ -636,7 +636,13 @@ namespace Maya.Util {
...
+                padding: 0 0.75em;
```
danrabbit: It would probably be better to either use `min-height` and `min-width` here or `height_request` and `width_request` in Gtk.

Something else to consider is using a Gtk.Grid instead of a Gtk.Label since Label will reserve extra space for text inside of it

cassidyjames:
@danrabbit I've switched to a grid and addressed the CSS comment.

PR #348, opened 2018-12-18
No spaces around en-dashes

### LINKED ISSUE #346: Typographically-correct ranges
We should be using an en-dash with no spaces for all date ranges.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67712605-typographically-correct-ranges?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Turns out we were already using en-dashes, but had spaces. Fixes #346

### DIFF
--- a/src/Widgets/AgendaEventRow.vala
+++ b/src/Widgets/AgendaEventRow.vala
@@ -188,7 +188,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
                 if (is_allday) {
                     datetime_string = "%s".printf (start_date_string);
                 } else {
-                    // TRANSLATORS: A range from start date and time to end time i.e. "Friday, Dec 21, 7:00 PM–9:00 PM"
+                    // TRANSLATORS: A range from start date and time to end time i.e. "Fri, Dec 21, 7:00 PM–9:00 PM"
                     datetime_string = _("%s, %s–%s").printf (start_date_string, start_time_string, end_time_string);
                 }
             }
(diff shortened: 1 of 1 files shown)


danrabbit:
I'm not sure I like having no spaces at all. This is what it looks like for reference:

screenshot from 2018-12-19 16 46 30 2x

cassidyjames:
@danrabbit yeah, that's typographically correct though. En dashes are never supposed to have spaces around them. I'm mostly indifferent about the look.

danrabbit:
Where is your reference for that? On the [Wikipedia style guide](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Dates_and_numbers#Ranges) for example I'm seeing that. "If at least one item on either side of the en dash contains a space, then a spaced en dash is used"

cassidyjames:
@danrabbit several random online grammar guides (and if I recall from school, Chicago and AP style, but I can't find a non-paywalled source for that). But I'm :+1: with following the Wikipedia style guide.

danrabbit:
Looks like that is AP style, but they also write "p.m." so they're weird. Without spaces, it makes "PM-8:20" look like the related letters instead of separating "7:20 PM" from "8:20 PM". I think no spaces looks :+1: when there are no spaces in what's on either side, so I think I'm with Wikipedia on this one.

PR #351, opened 2018-12-19
AgendaEventRow: Search OpenStreetMap for locations

danrabbit:
iOS calendar adds `\n` for all location strings, so we can easily get just the address by looking for this and splitting there. If there's no split, we just search for the whole string.


cassidyjames:
Works sometimes on my events. One thing I noticed is that it looks like we're not URL encoding the strings.

jeremypw:
This works pretty well for me with events created in elementary calendar at least.  However, there is no tooltip when hovering over the location text indicating what will happen if you click on it (and the text is not underlined or otherwise indicating that it is a hyperlink to a map).

cassidyjames:
@jeremypw no it should probably be more obviously a link without requiring mystery-meat hovering around.

jeremypw:
@danrabbit Converting to draft as has uaddressed issues and no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

danrabbit:
Yeah this is kind of doo doo, closing

PR #356, opened 2018-12-27
CalendarButton: Add a simple search

### LINKED ISSUE #214: Select event's Calendar using keyboard?
When entering a new event I would love to be able to select the target Calendar using just my keyboard. Especially since I just typed the title and I'll might need to type again to enter the time for example. 

On the web the convention for dropdown menu's seems to be that typing the desired value results in that value being automatically selected. It's also possible to traverse all the values using the arrow keys. 

Relevant Widget: [CalendarButton](https://github.com/elementary/calendar/blob/master/src/Widgets/CalendarButton.vala)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/53521551-select-event-s-calendar-using-keyboard?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #214 

screenshot from 2018-12-27 13 16 36 2x


danrabbit:
@jeremypw Okay I think I have most of those issues solved. I'm not sure how to move focus back to the search entry when we reach the top of the list though. Any tips/ideas there?

Something that might be helpful to note here for future reference is that listbox seems to retain the selection even for hidden widgets. So when focus was moved into the listbox, it first went to a hidden listboxrow and then only after navigating a second time was it moved to a visible row. Neat.

jeremypw:
@danrabbit : Here is one way:

```
        list_box.move_cursor.connect ((mvmt, count) => {
            var row = list_box.get_selected_row ().get_index ();
            if (row == 0 && count == -1) {
                search_entry.grab_focus ();
            }
        });
```

danrabbit:
@jeremypw oh duh, that looks good and works great. Thanks! :)

PR #358, opened 2018-12-29
ReminderPanel: misc cleanup

danrabbit:
* Update copyright header
* Add margin to the grid instead of each child
* Use style classes instead of sensitivity for styling label
* Remove dead code
* Organize
* Change labels ("0 minutes" → "At time of event")
* Expand combobox for larger click target
* Destructive action class on delete button


src/EventEdition/ReminderPanel.vala
```diff
@@ -1,6 +1,5 @@
-// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/*-
- * Copyright (c) 2011-2015 Maya Developers (http://launchpad.net/maya)
+/*
+ * Copyright 2011-2018 elementary, Inc. (https://elementary.io)
```
jeremypw: Since this will be merged in 2019 you may as well copyright it till then,

PR #372, opened 2019-01-07
Application: Don't init clutter

danrabbit:
This app doesn't use clutter


tintou:
It does, the Map won't work if you don't initialize clutter

danrabbit:
TIL

PR #379, opened 2019-02-15
ContractorButtonWithMenu: Use FileChooserNative

danrabbit:


src/Widgets/ContractorButtonWithMenu.vala
```diff
@@ -79,23 +79,22 @@ namespace Maya.View.Widgets {
...
+            filter.add_mime_type ("text/calendar");
+
+            filechooser = new Gtk.FileChooserNative (
+                _("Export Calendar…"),
+                null,
+                Gtk.FileChooserAction.SAVE,
+                _("Save"),
+                _("Cancel")
+            );
+            filechooser.do_overwrite_confirmation = true;
```
jeremypw: Not supported across portal ?? (to be confirmed).

danrabbit: Yeah, the new native chooser should probably just do this and not provide an option. This is only for the fallback chooser

PR #386, opened 2019-03-29
Set color_button_red active when creating a new calendar

ryonakano:
### BEFORE

2019-03-29 09 35 09 の画面録画

When you create a new calendar, the red radiobutton (`color_button_red`) is selected for the color of calendar by default. However, once you try to edit existing calendars ("Personal" or "Birthdays & Anniversaries"), no radiobutton is selected (actually an hidden raddiobutton, `color_button_none`, is selected because the color of these calendars is not listed in the default color palette). And then if you try to create a new calendar, the red radiobutton is no longer selected by default, since the hidden radiobutton is still selected.

### AFTER

2019-03-29 09 36 51 の画面録画

This PR explicitly select the red radiobutton when users try to create a new calendar.


danrabbit:
Thanks for these fixes!

PR #401, opened 2019-06-02
Get the time format from the new key and make it more efficient

tintou:


peteruithoven:
Looks like we'd need to use the same logic to set the format here:
https://github.com/elementary/calendar/blob/0087c1800d72be525e00e186b0fcae646f40b4a7/src/EventEdition/InfoPanel.vala#L276-L281

jeremypw:
@tintou Converting to draft as some outstanding issues and conflicts and no recent activity. Please feel free to close or reopen for review after fixing as required.

PR #403, opened 2019-06-09
Remove unused po

cjfloss:
- Remove unused po files
- Update pot files
- Update po files


tintou:
We have to keep them otherwise they won't show-up in Weblate

cjfloss:
I saw the project [Remove Unused Translations](https://github.com/orgs/elementary/projects/51) and thought it would be applied to other projects as well, since the unused po files were removed from calculator and appcenter

PR #416, opened 2019-08-12
Use keyword lists for event icons

jeremypw:
Based on #322 

Refactored to use keyword lists and hashmaps. 

Also adds translator comments and reduces added line count.


jeremypw:
There is a problem with false hits, for example the keywords "bar" (DRINKS category) and "barbecue" (FOOD category).  A simplistic match would find "bar" in an event named "barbecue" and wrongly classify it as "DRINKS" instead of food.  For some languages, putting spaces around the keywords would prevent this, but this would not work in other languages.  In this PR, matches on longer keywords are weighted, which fixes the bar/barbecue issue but this may not work in all situations.

src/Widgets/AgendaEventRow.vala
```diff
@@ -35,12 +35,61 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+        CELEBRATION,
+        CALL,
+        DRINKS,
+        DRIVING,
+        FLIGHT,
+        FOOD,
+        LEGAL,
+        MOVIE,
+        WEDDING,
+        N_CATEGORIES
```
tintou:
```suggestion
        N_CATEGORIES;
        public unowned string get_builtin_keywords () {
            switch (this) {
                …
            }
        }
```
Doing this would probably simplify the code afterwards

jeremypw: Thanks, that is much better.  I have also done the same for the icon name since there is only one icon per category,

jeremypw:
Other categories might be desirable e.g. "MEDICAL" but that is left for another PR.

src/Widgets/AgendaEventRow.vala
```diff
@@ -213,4 +361,13 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
             critical (e.message);
         }
     }
+
+    private void split_keywords (Category category) {
+        var words = category.get_builtin_keywords ().split (";", 20);
```
tintou: Is there any specific reason to set a limit?
```suggestion
        var words = category.get_builtin_keywords ().split (";");
```

src/Widgets/AgendaEventRow.vala
```diff
@@ -213,4 +361,13 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+        foreach (string? word in words) {
```
tintou:
```suggestion
        foreach (unowned string? word in words) {
```

src/Widgets/AgendaEventRow.vala
```diff
@@ -35,12 +35,126 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+                case WEDDING:
+                    return "event-wedding-symbolic";
+
+                default:
+                    return null;
+            }
+        }
+    }
+
+//    private Gee.HashMap<Category, string> category_icon_map;
```
tintou: this line can go away

src/Widgets/AgendaEventRow.vala
```diff
@@ -35,12 +35,126 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
+    private Gee.HashMultiMap<Category, string> keyword_map; /* Would TreeMultiMap be better? */
```
tintou: You probably don't want to leave this question remaining forever :)
```suggestion
    private Gee.HashMultiMap<Category, string> keyword_map;
```
As there is no redundancy in key values, a HashMultiMap seems fine

src/Widgets/AgendaEventRow.vala
jeremypw: No problem envisaged if this removed.

src/Widgets/AgendaEventRow.vala
jeremypw: Done

src/Widgets/AgendaEventRow.vala
jeremypw: OOps

src/Widgets/AgendaEventRow.vala
jeremypw: Removed comment

PR #456, opened 2019-11-14
Reset AgendaView items when parameters change

### LINKED ISSUE #452: Changing dates repeatedly adds events into agenda sidebar
It looks like events are being added under "Next Week", "Next Month", etc. but never being removed or deduplicated. If you scroll around the calendar a bit with some events, eventually you get a ton of duplicated events in the sidebar.

Screenshot from 2019-10-29 15 56 54

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/82847047-changing-dates-repeatedly-adds-events-into-agenda-sidebar?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Thalexander:
Fixes #452 

When the month or year is changed on the calendar the new parameters cause the source to be reloaded, this will trigger on_events_added but not on_events_removed. To handle this I have connected to parameters_changed and delete existing rows, this is similar to the handling in the CalendarView although that runs on idle removing events and then syncing with the model, such a implementation here would require quite complex handling here to avoid duplication from the subsequent on_events_added signal.

### DIFF
--- a/src/AgendaView.vala
+++ b/src/AgendaView.vala
@@ -359,11 +359,11 @@ public class Maya.View.AgendaView : Gtk.ScrolledWindow {
         GLib.List<weak Gtk.Widget> selected_date_events_children = selected_date_events_list.get_children ();
         GLib.List<weak Gtk.Widget> upcoming_events_children = upcoming_events_list.get_children ();
 
-        foreach (var row in selected_date_events_children) {
+        foreach (unowned Gtk.Widget row in selected_date_events_children) {
             row.destroy ();
         }
 
-        foreach (var row in upcoming_events_children) {
+        foreach (unowned Gtk.Widget row in upcoming_events_children) {
             row.destroy ();
         }
     }
--- a/src/AgendaView.vala
+++ b/src/AgendaView.vala
@@ -111,6 +111,7 @@ public class Maya.View.AgendaView : Gtk.ScrolledWindow {
         calmodel.events_added.connect (on_events_added);
         calmodel.events_removed.connect (on_events_removed);
         calmodel.events_updated.connect (on_events_updated);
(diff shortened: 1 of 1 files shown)


Thalexander:
This will remove the upcoming events not visible on the calendar though, although this is as a result of them not loading unless they are withing the calendar models parameters, perhaps the upcoming events list should perform differently. Selected events list works as expected.

src/AgendaView.vala
```diff
@@ -351,6 +352,22 @@ public class Maya.View.AgendaView : Gtk.ScrolledWindow {
...
     }
 
+    /**
+     * Calendar model parameters have been updated.
+     */
+    private void on_model_parameters_changed () {
+        GLib.List<weak Gtk.Widget> selected_date_events_children = selected_date_events_list.get_children ();
+        GLib.List<weak Gtk.Widget> upcoming_events_children = upcoming_events_list.get_children ();
+
+        foreach (var row in selected_date_events_children) {
```
tintou:
```suggestion
        foreach (unowned Gtk.Widget row in selected_date_events_children) {
```

src/AgendaView.vala
```diff
@@ -351,6 +352,22 @@ public class Maya.View.AgendaView : Gtk.ScrolledWindow {
...
+            row.destroy ();
+        }
+
+        foreach (var row in upcoming_events_children) {
```
tintou:
```suggestion
        foreach (unowned Gtk.Widget row in upcoming_events_children) {
```

PR #458, opened 2019-11-17
Stop more label becoming white when selected

### LINKED ISSUE #399: More text is hidden when day is selected
I don't understand why the "... more..." text is hidden when a day is selected?
Screen record from 2019-05-23 18 21 00

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74559078-more-text-is-hidden-when-day-is-selected?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Thalexander:
Fixes #399 

The more label was set to white by the elementary stylesheet when selected, this sets text colour for cell as @fg_color and so it stays visible when selected.

### DIFF
--- a/data/style/Grid.css
+++ b/data/style/Grid.css
@@ -20,6 +20,7 @@
     border-radius: 0;
     border-top: solid 1px @menu_separator;
     border-left: solid 1px @menu_separator;
+    color: @fg_color;
 }
 
 .cell:dir(rtl) {


4jNsY6fCVqZv:
Do you think a solution should be implemented locally on app level or globally on stylesheet level?

Thalexander:
This is a local issue, we are overwriting the stylesheet that would have changed the background colour, this then causes the issue where the text becomes invisible as the stylesheet is still controlling the text colour and assuming the background colour has changed from white.

4jNsY6fCVqZv:
Thanks for the explanation and of course for fixing the issue!

PR #476, opened 2019-12-15
Add "Duplicate" to the context menu

### LINKED ISSUE #475: Duplicate calendar item using the context menu
## Prerequisites

## Feature
**Is your feature request related to a problem? Please describe.**

It's not possible to create an exact duplicate of an existing calendar entry. For example, I manage an event calendar for a local user group; calendar items repeat every four weeks and are all similar (same headline, same URL, etc.), but Calendar doesn't support me in any way in creating a new calendar item that's identical to an existing one.

**Describe the solution you'd like**

Add "Duplicate" to the context menu when right-clicking a calendar item (which currently contains "Remove" and "Edit"). Creates and saves a new calendar item identical to the one that was right-clicked (no intermediate dialog). The user can then double-click one of the two items and edit it.

**Existing work**

The macOS calendar application has been doing it this way for years.

**Describe alternatives you've considered**

Since Elementary Calendar doesn't even seem to support copy/paste of complete calendar items, the only alternative I can think of is to use a different calendar program.

**Additional context**

Screenshot of the context menu in macOS.

Bildschirmfoto 2019-12-15 um 12 10 46

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/85486521-duplicate-calendar-item-using-the-context-menu?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wolframroesler:
Creates an exact duplicate of an existing calendar item (same contents, same calendar). The same feature exists  in the macOS Calendar application.

Screenshot from 2019-12-15 17 20 30

Fixes #475

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -137,6 +137,26 @@ public class Maya.MainWindow : Gtk.ApplicationWindow {
         }
     }
 
+    public void on_duplicate (ECal.Component comp) {
+        E.Source src = comp.get_data ("source");
+
+        if (src.writable == true && Model.CalendarModel.get_default ().calclient_is_readonly (src) == false) {
+            // The event editor dialog (EventDialog) uses its date/time parameter to tell
+            // if we're editing an existing event (parameter is null) or creating a new one
+            // (parameter is not null). Since here we're creating a new event as a copy of
+            // an existing one, we have to pass that event's date/time.
+            DateTime from_date, _;
+            Util.get_local_datetimes_from_icalcomponent (comp.get_icalcomponent (), out from_date, out _);
+
+            // Now open the editor dialog.
+            var dialog = new Maya.View.EventDialog (comp, from_date);
+            dialog.transient_for = this;
+            dialog.present ();
+        } else {
+            Gdk.beep ();
+        }
+    }
+
     public override bool configure_event (Gdk.EventConfigure event) {
         if (configure_id != 0) {
             GLib.Source.remove (configure_id);
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -140,7 +140,7 @@ public class Maya.MainWindow : Gtk.ApplicationWindow {
     public void on_duplicate (ECal.Component comp) {
         E.Source src = comp.get_data ("source");
 
-        if (src.writable == true && Model.CalendarModel.get_default ().calclient_is_readonly (src) == false) {
+        if (src.writable && ! Model.CalendarModel.get_default ().calclient_is_readonly (src)) {
             // The event editor dialog (EventDialog) uses its date/time parameter to tell
             // if we're editing an existing event (parameter is null) or creating a new one
             // (parameter is not null). Since here we're creating a new event as a copy of
--- a/src/Widgets/EventMenu.vala
+++ b/src/Widgets/EventMenu.vala
@@ -5,12 +5,12 @@
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
@@ -29,7 +29,7 @@ public class Maya.EventMenu : Gtk.Menu {
 
     construct {
         E.Source src = comp.get_data ("source");
-        bool can_modify_cal = Model.CalendarModel.get_default ().calclient_is_readonly (src) == false;
+        bool can_modify_cal = ! Model.CalendarModel.get_default ().calclient_is_readonly (src);
         bool sensitive = src.writable && can_modify_cal;
 
         var edit_item = new Gtk.MenuItem.with_label (_("Edit…"));
--- a/src/Widgets/EventMenu.vala
+++ b/src/Widgets/EventMenu.vala
@@ -51,7 +51,7 @@ public class Maya.EventMenu : Gtk.Menu {
         remove_item.sensitive = sensitive;
         remove_item.activate.connect (remove_event);
 
-        var dup_item = new Gtk.MenuItem.with_label (_("Duplicate"));
+        var dup_item = new Gtk.MenuItem.with_label (_("Duplicate…"));
         dup_item.sensitive = can_modify_cal;
 
         append (remove_item);
@@ -80,9 +80,8 @@ public class Maya.EventMenu : Gtk.Menu {
         var dup = comp;
         dup.set_uid (uid);
 
-        // Store the duplicated event in the calendar
-        var calmodel = Model.CalendarModel.get_default ();
-        calmodel.add_event (comp.get_data<E.Source> ("source"), dup);
+        // Open the new event in the editor
+        ((Maya.Application) GLib.Application.get_default ()).window.on_duplicate (dup);
     }
 
     private void add_exception () {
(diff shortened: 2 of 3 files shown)


wolframroesler:
Hello, is anyone reading this? Like it? Hate it? Approve and merge it? Thanks :)

jeremypw:
I guess if MacOS has it, it must be a good idea (:-þ)  However, I have asked a design team member to confirm.

src/Widgets/EventMenu.vala
```diff
@@ -29,7 +29,8 @@ public class Maya.EventMenu : Gtk.Menu {
 
     construct {
         E.Source src = comp.get_data ("source");
-        bool sensitive = src.writable == true && Model.CalendarModel.get_default ().calclient_is_readonly (src) == false;
+        bool can_modify_cal = Model.CalendarModel.get_default ().calclient_is_readonly (src) == false;
```
jeremypw: Suggest use `bool can_modify_cal = !Model.CalendarModel.get_default ().calclient_is_readonly (src);`

jeremypw:
I would think that the duplicate event should be immediately opened in the event editor since, presumably the user will not want to absolutely identical events.

wolframroesler:
@jeremypw You're right, I never use "duplicate" without a double-click immediately afterwards. Maybe another question for the design team? Anyway, I implemented it the way that macOS does it (which doesn't necessarily mean that it's the best way).

src/Widgets/EventMenu.vala
wolframroesler: Of course. "==false" is plain stupid. Will change before anyone sees it (I keep flagging stuff like that in code reviews all the time.)

wolframroesler:
Hi @jeremypw, any decision from the design team yet?

lainsce:
I'd make this duplicate the event and open the event editor dialog instantly for it instead of just duplicating the event as discussed earlier, and the label would become `Duplicate Event…` by then, my 2¢.

cassidyjames:
Yeah maybe `Duplicate and Edit…` would be the most precise? I agree that immediately opening the event edit dialog makes sense, because you're basically using an existing event as a template to make a new event.

wolframroesler:
I find the ancient convention of putting "..." after the menu text whenever the menu item opens a dialog ugly and outdated, but that's just my personal opinion. But, since "Edit ..." follows that convention, "Duplicate" should also follow it, so how about "Duplicate ..."?

Or, let's just decide to do away with 1987's CUA and change "Edit ..." to "Edit" because nobody's going to miss the ellipsis anyway. Or am I being too radical here?

lainsce:
The elementary OS Human Interface Guidelines states, and I quote:
"An ellipsis should be used to let a user know that more information or a further action is required before their action can be performed. Usually this means that the user should expect a new interface element to appear such as a new window, dialog, toolbar, etc in which they must enter more information or make a selection before completing the intended action."

So it's not about CUA, but actually cohesion with the whole system.

cassidyjames:
My concern with just `Duplicate…` is that it may not be immediately clear why duplicating will require further action, whereas `Duplicate and Edit…` seems clearer to me. But just `Duplicate…` might be fine.

lainsce:
In my mindset, when I'm duplicating some event I already expect that I'll be able to edit my duplicate before ending the chain of actions. To me, "Edit" is implicit...

wolframroesler:
Changed menu text to "Duplicate…". Now invoking editing dialog for the new entry (in "add new item" mode, i. e. when the user cancels the editor, the new event is discarded).

cassidyjames:
This seems to work as expected with one exception: if I am duplicating a recurring event, it seems that I cannot enable recurring events on the new (duplicate) event.

jeremypw:
Should the duplicate of recurring event be recurring by default? Or should the user reset this?

wolframroesler:
I'm having trouble changing an existing event to recurring even without duplicating it. Could it be that we're dealing with bugs that are independent of my changes? This comment https://github.com/elementary/calendar/issues/483#issuecomment-611685319 is hinting in that direction.

wolframroesler:
> Should the duplicate of recurring event be recurring by default?

Yes, the duplicate should be an exact copy unless the user changes it.

src/MainWindow.vala
```diff
@@ -137,6 +137,26 @@ public class Maya.MainWindow : Gtk.ApplicationWindow {
         }
     }
 
+    public void on_duplicate (ECal.Component comp) {
+        E.Source src = comp.get_data ("source");
+
+        if (src.writable == true && Model.CalendarModel.get_default ().calclient_is_readonly (src) == false) {
```
jeremypw, line 143: `== true` can be omitted.  `== false` can be replaced by `!(....)`

src/MainWindow.vala
```diff
@@ -137,6 +137,26 @@ public class Maya.MainWindow : Gtk.ApplicationWindow {
...
+            // The event editor dialog (EventDialog) uses its date/time parameter to tell
+            // if we're editing an existing event (parameter is null) or creating a new one
+            // (parameter is not null). Since here we're creating a new event as a copy of
+            // an existing one, we have to pass that event's date/time.
+            DateTime from_date, _;
+            Util.get_local_datetimes_from_icalcomponent (comp.get_icalcomponent (), out from_date, out _);
+
+            // Now open the editor dialog.
+            var dialog = new Maya.View.EventDialog (comp, from_date);
```
jeremypw, line 152: The dialog needs to be amended so that it can distinguish between creating an new event, editing an existing event  and duplicating an existing event.  At the moment it does not do this properly and therefore does not duplicate recurring events (just create a new standard event at the same time)

jeremypw:
@wolframroesler Are you able to continue work on this PR, in particular the problem duplicating recurring events?

wolframroesler:
Will try, however I'm afraid that problem isn't related to my changes, since I'm having trouble changing an existing event to recurring even without duplicating it.

jeremypw:
@wolframroesler OK, thanks.  Let me know if you would prefer me or someone else to look into it - I wasn't sure whether you were still working on it.

wolframroesler:
@jeremypw please have a look because it's behaving weirdly: The duplicate shows up only once but definitely has the "repeating" flag set.

calender-duplicate-repeating

Duplicating a non-repeating entry and then setting it to "repeating seems to work fine ...

calendar-duplicate

... however changing an existing non-repeating entry to repeating also has problems.

calendar-repeating

Maybe you can try and fix the last one (which doesn't involve "Duplicate"), and then we'll see if and how "Duplicate" behaves.

jeremypw:
OK, I'll have a look over the next day or so.

jeremypw:
I haved pushed a PR #516 that makes some progress towards adding and removing recurrences. Needs more work for editing existing occurrences.

jeremypw:
@wolframroesler I have now pushed a new branch fixing issues in editing events and a further branch based on it that implements event duplication. You were right that several serious underlying problems needed fixing before it was possible to implement event duplication!  If you have a chance to try these branches out I would welcome your comments

wolframroesler:
@jeremypw I'm afraid I won't be putting more time into this as I have essentially given up on the native elementaryOS apps, which (despite being stunningly beautiful to behold) have turned out to be too buggy for me to trust them during everyday use. I gave Calendar, Mail etc. more than one chance, but I finally got too frustrated and switched to alternatives, which aren't nearly as attractive but which get the job done in a much more trustworthy way. I'd love to stand corrected, and I'm hoping to return once this project has gained more maturity. IMO elementaryOS has a lot of potential, its UI being almost on par with macOS and its semi-commercial distribution model being innovative and promising, but I really require a rock-solid core underneath the pretty interface.

Feel free to use the code I contributed so far.

jeremypw:
@wolframroesler That's a pity but thank you for your contributions.  I understand what you are saying about the elementary native apps.  The more experienced/talented developers there are who are willing to work on the nitty gritty of these apps the sooner the standard their function will match that of their design.

jeremypw:
Closing in favour of #520

PR #482, opened 2020-01-05
Don't overwrite events unnecessarily when parsing

### LINKED ISSUE #453: Editing the title of an event reschedules it to today's next whole hour
Looks to be a regression from #209. Might require pressing Enter in the event name when editing, I haven't fully tracked down when it happens but it's fairly reproducible.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/82847129-editing-the-title-of-an-event-reschedules-it-to-today-s-next-whole-hour?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Thalexander:
Fixes #453 

Adds a date_parsed and time_parsed property to determine whether to overwrite existing data. This may not handle the phrase "for x minutes/hours" (and equivalent German) correctly, however this was already not correct and so I would consider it a separate issue, we will need to determine if the date changes when saying a length that goes over the current day, I kept the existing behaviour of wiping data for these cases. Does not implement guest parsing.


jeremypw:
It seems a bit bizarre that there has to be special code for particular languages hard-coded in just to edit the title of an event.  I haven't looked at the code in detail but is this really necessary?

Thalexander:
It is necessary to convey whether a time or a date has been detected, if we were to handle this in ParsedEvent I believe the only real way of doing so would be to have two separate DateTimes and then we would need to change the languages to not assign a DateTime by default but only if parsed, and then this in turn might complicate interactions when multiple expressions are parsed.

jeremypw:
@Thalexander I cannot seem to reproduce the issue in master so it is difficult to check this fixes it.  Do you have a way of reproducing the issue?  I have tried with both US and UK english locales.

PR #486, opened 2020-01-22
Avoid repating event time if end equals start

### LINKED ISSUE #463: If start and end time are the same, don't show both
Screenshot from 2019-11-26 10 38 54@2x

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/84449676-if-start-and-end-time-are-the-same-don-t-show-both?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

AlbertAlonso:
Fixes #463

Small fix to avoid repeating the same time for an event. However, the app didn't allow to create events with the same time in the start and in the end, so a small change was made there too.

### DIFF
--- a/src/Widgets/AgendaEventRow.vala
+++ b/src/Widgets/AgendaEventRow.vala
@@ -300,7 +300,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
         string end_time_string = end_date.format (Settings.time_format ());
         string? datetime_string = null;
 
-        var is_same_time = start_date_string == end_date_string;
+        var is_same_time = start_time_string == end_time_string;
 
         datatime_label.show ();
         datatime_label.no_show_all = false;
--- a/src/Widgets/AgendaEventRow.vala
+++ b/src/Widgets/AgendaEventRow.vala
@@ -33,6 +33,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
     public string summary { public get; private set; }
     public bool is_allday { public get; private set; default = false; }
     public bool is_multiday { public get; private set; default = false; }
+    public bool is_same_time { public get; private set; default = false; }
     public Gtk.Revealer revealer { public get; private set; }
 
     private Gtk.Image event_image;
@@ -318,7 +319,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
                 } else {
                     if (is_same_time){
                         // TRANSLATORS: A single time i.e. "7:00 PM"
-                        datetime_string = C_("%s").printf (start_time_string);
+                        datetime_string = _("%s").printf (start_time_string);
                     } else {
                         // TRANSLATORS: A range from start time to end time i.e. "7:00 PM – 9:00 PM"
                         datetime_string = C_("time-range", "%s – %s").printf (start_time_string, end_time_string);
@@ -330,7 +331,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
                 } else {
                     if (is_same_time){
                         // TRANSLATORS: A single time from the start date i.e. "Friday, Dec 21, 7"00 PM"
-                        datetime_string = C_("%s, %s").printf (start_date_string, start_time_string);
+                        datetime_string = _("%s, %s").printf (start_date_string, start_time_string);
                     } else {
                         // TRANSLATORS: A range from start date and time to end time i.e. "Friday, Dec 21, 7:00 PM – 9:00 PM"
                         datetime_string = _("%s, %s – %s").printf (start_date_string, start_time_string, end_time_string);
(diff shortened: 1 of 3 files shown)


src/Widgets/AgendaEventRow.vala
```diff
@@ -315,15 +317,25 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
...
                 }
             } else {
                 if (is_allday) {
                     datetime_string = "%s".printf (start_date_string);
                 } else {
-                    // TRANSLATORS: A range from start date and time to end time i.e. "Friday, Dec 21, 7:00 PM – 9:00 PM"
-                    datetime_string = _("%s, %s – %s").printf (start_date_string, start_time_string, end_time_string);
+                    if (is_same_time) {
+                        // TRANSLATORS: A single time from the start date i.e. "Friday, Dec 21, 7"00 PM"
+                        datetime_string = _("%s, %s").printf (start_date_string, start_time_string);
```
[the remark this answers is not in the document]
cassidyjames: Sounds sane to me since we don't have a range to display.

src/Widgets/AgendaEventRow.vala
```diff
@@ -315,15 +317,25 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
                     datatime_label.hide ();
                     datatime_label.no_show_all = true;
                 } else {
-                    // TRANSLATORS: A range from start time to end time i.e. "7:00 PM – 9:00 PM"
-                    datetime_string = C_("time-range", "%s – %s").printf (start_time_string, end_time_string);
+                    if (is_same_time) {
+                        // TRANSLATORS: A single time i.e. "7:00 PM"
+                        datetime_string = _("%s").printf (start_time_string);
```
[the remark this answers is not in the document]
AlbertAlonso: Totally right! Should I change the 
```
if (is_allday) {
    datetime_string = "%s".printf (start_date_string);
}
```
a few lines below?!

src/Widgets/AgendaEventRow.vala
```diff
@@ -291,6 +292,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
 
         is_allday = Util.is_all_day (start_date, end_date);
         is_multiday = Util.is_multiday_event (ical_event);
+        is_same_time = Util.is_same_time (start_date, end_date);
```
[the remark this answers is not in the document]
AlbertAlonso: Yeah, my line of thought was that you may need to add some checks or wanted to use it somewhere else. But I think make it clear for now is better. Thanks!

AlbertAlonso:
@danrabbit Thank you for the feedback! I think I have implemented all the suggestions.

src/Widgets/AgendaEventRow.vala
```diff
@@ -291,6 +292,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
 
         is_allday = Util.is_all_day (start_date, end_date);
         is_multiday = Util.is_multiday_event (ical_event);
+        var is_same_time = start_date.equals (end_date);
```
danrabbit: Looks like we accidentally added an `s` here

```suggestion
        var is_same_time = start_date.equal (end_date);
```

AlbertAlonso: Damn it!

src/Widgets/AgendaEventRow.vala
```diff
@@ -315,15 +317,24 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
                     datatime_label.hide ();
                     datatime_label.no_show_all = true;
                 } else {
-                    // TRANSLATORS: A range from start time to end time i.e. "7:00 PM – 9:00 PM"
-                    datetime_string = C_("time-range", "%s – %s").printf (start_time_string, end_time_string);
+                    if (is_same_time) {
```
danrabbit: So I actually have a webcal that has an event which is a deadline and for some reason even with this branch it's reading as `4:00 pm – 4:00pm`. I wonder if we should compare the start and end time strings and not the actual datetimes. I wonder if the end datetime has like a null date or something weird

AlbertAlonso: I changed it to compare strings and it works on my previous events. :crossed_fingers:

src/Widgets/AgendaEventRow.vala
```diff
@@ -299,6 +300,8 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
         string end_time_string = end_date.format (Settings.time_format ());
         string? datetime_string = null;
 
+        var is_same_time = start_date_string == end_date_string;
```
danrabbit: Should be time string, not date string ;)

AlbertAlonso: Hahaha I'm so sorry!! Not my best day

danrabbit: Happens to us all!

PR #499, opened 2020-04-01
Use GLib.Settings for saved state

### LINKED ISSUE #440: Replace Granite.Services.Settings with GLib.Settings
<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/82324446-replace-granite-services-settings-with-glib-settings?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #440


core/Model/CalendarModel.vala
```diff
@@ -60,20 +60,25 @@ public class Maya.Model.CalendarModel : Object {
...
+    private static GLib.Settings state_settings;
 
     public static CalendarModel get_default () {
         if (calendar_model == null)
             calendar_model = new CalendarModel ();
         return calendar_model;
     }
 
+    static construct {
+        state_settings = new GLib.Settings ("io.elementary.calendar.savedstate");
```
jeremypw, line 72: This is the same as `Application.saved_state`. Do we need a second instance?

danrabbit: Yeah this is in the library though so I don’t think it has access to the application

jeremypw: Ah, OK. Although presumable the application has access to the library so we may still not need two instances?

danrabbit: Yeah that’s valid. Could get things from calendarmodel instead

jeremypw: Can be done in another PR

Issue #267, opened 2018-09-11
If dark mode is preferred, the Calendar UI is almost unusable


phase1geo:
Even though dark mode is currently not officially supported, it is the only prebuilt application that looks terrible and becomes almost unusable when dark mode preferred is set. Additionally, if other themes besides the stock elementary theme is applied, it also becomes functionally unusable.

eljefuri:
@phase1geo Sorry to immediately take this off-topic, but do you know if dark mode is something that the Elementary developers have any plans to implement as a mainline feature? If so there is absolutely an argument for fixing this across any apps that have issues in dark mode.

phase1geo:
I don't think that the elementary OS developers have publicly stated support for a system-wide dark mode. In fact, I believe that they have argued against having system-wide support for dark mode.

thomsch98:
My personal "solution" to the same problem was to clone the source code and to drop the calendar-coloring. 

It only was necessary to comment out one line in EventButton.vala and one from AgendaEventRow.vala:

EventButton.vala ~ 50
```
grid_style_context = internal_grid.get_style_context ();
        grid_style_context.add_class ("event");
//       grid_style_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);

        var event_box = new Gtk.EventBox ();
        event_box.events |= Gdk.EventMask.BUTTON_PRESS_MASK;
```
AgendaEventRow.vala: ~ 100
```
        main_grid_context = main_grid.get_style_context ();
        main_grid_context.add_class ("event");
//       main_grid_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);

        var event_box = new Gtk.EventBox ();
        event_box.add (main_grid);
```

Now I've got myself a usable calendar ....

![image](https://user-images.githubusercontent.com/2971345/54428340-83a89d00-471d-11e9-9b76-c7c6f88bcbed.png)

phase1geo:
That looks pretty nice! Thanks for the tip!

DenisPalchuk:
From my point of view it's just a workaround. @thomsch98 just suggest to disable coloring at all.
The real problem is wrong alfa value when we define background for events. I just made a PR with changes and right it looks really better for me:
![image](https://user-images.githubusercontent.com/3480522/78977349-a5db3d00-7b20-11ea-8bd2-0888fec9ce25.png)

Link to PR #504

PR #504, opened 2020-04-10
Fix tales background for better visibility in dark mode

DenisPalchuk:
I know that dark mode is a hidden feature but I really believe that anyway design team will face with this issue later. So I just wanna help a little. This PR just increase alfa param to accent_color in events background. It increase readability not only for dark mode but for light mode as well
Before:
image
After:
image


DenisPalchuk:
Fixes #267

Issue #491, opened 2020-02-16
Selecting 'Remove Occurence" from context menu crashes the Calendar


Andrewsudarikov:
## Describe the bug
Clicking on the context menu item "Remove occurrence" immediately crashes the program. 

### To Reproduce
1. Create an event. 
2. RIght-click on a new event. 
3. Select "Remove occurence"
4. Voila, the program reliably crashes every time. 

### Expected behavior
An event occurrence for the selected date should be removed. The application must continue running. 

## Platform Information
 - OS: Elementary OS 5
 - OS Version 5.1.2

- [X] I'm using the latest released stable version

mcclurgm:
I've been seeing this too. I've found a stack trace for the issue. It's in `EventMenu.vala` line 80. (Following a stack trace from valgrind) It appears that the issue is with setting the exception dates in the call to the Evolution server. Specifically, it's the call to `e_cal_component_set_exdate_list`. 

I'm not sure if this is helpful, but this behavior doesn't appear when adding exceptions in the editor because that uses libical properties, not libecal. Maybe that's helpful to see where this behavior is coming from, I'm not familiar with Evolution.

mkykadir:
I think problem relies on line 77 in `EventMenu.vala`, I checked Vala compiler output for `EventMenu.vala` file which is `EventMenu.c` and problem can be seen that `dtstart` is type of `ECalComponentDateTime*` and `e_cal_component_get_dtstart` function receives address of `dtstart`. So it sends address of the pointer which is wrong.

```C	
ECalComponentDateTime* dtstart = NULL;
// ...
ECalComponentDateTime* _tmp1_ = NULL; // this will be dtstart
// ...
#line 68 "/home/mkytr/Projects/calendar/src/Widgets/EventMenu.vala"
g_return_if_fail (self != NULL);
#line 77 "/home/mkytr/Projects/calendar/src/Widgets/EventMenu.vala"
_tmp0_ = self->priv->_comp;
#line 77 "/home/mkytr/Projects/calendar/src/Widgets/EventMenu.vala"
e_cal_component_get_dtstart (_tmp0_, &_tmp1_); // __tmp1__ becomes ECalComponentDateTime**
```
Because of that invalid address is updated so inner library functions gets invalid memory which it cannot access.
[...]

PR #505, opened 2020-04-10
Make repeat switch active when required

mcclurgm:
Solves an issue that @mkykadir found ([original](https://github.com/elementary/calendar/issues/491#issuecomment-612106879) from #491):
> Also anyone noticed that edit panel doesn't enables "Repeat" toggle even repeat is enabled for the event? I am also looking solution for that.

In the constructor for `RepeatView`, the `load` function was called out of order and it was leading to some issues where data would be overwritten or not dealt with properly. Specifically in this case, `load` was called (line 245) before `repeat_switch` got an event handler and before it was set to inactive (line 253). This meant that the `load` function both wasn't able to update the active state of any other widget in the window to be active, and whatever state it set was overwritten later anyway. This PR fixes that by simply moving the call to `load` after all the setup for `repeat_switch` is complete.

A small design note, I decided to separate the `load` call from attaching all the widgets to the window grid, which is how the code was organized before. I'm happy to move those `attach` calls down too, to preserve the original organization.


mkykadir:
This solves the problem, switch is active and all components below it is active and editable

PR #506, opened 2020-04-11
Fix behavior of removing occurrences of recurrent events

### LINKED ISSUE #491: Selecting 'Remove Occurence" from context menu crashes the Calendar

mkykadir:
Fixes #491

VAPI for `libecal-1.2` was wrong that caused to send address of the pointer instead of the pointer itself. Program was crashing with _Segmentation fault_ due to invalid access to an address. Problem was explained in detail at [this comment](https://github.com/elementary/calendar/issues/491#issuecomment-611752958) of the issue report. Solving VAPI problem revealed another problem which was more dangerous than _Segmentation fault_ error of the problem, that may broke down event of the user. When user selects _Remove occurrence_ menu item, misbehavior occurs and only right-clicked event remains and other occurrences will be deleted. I found that instead of calling `update_event` method to update whole event by adding exception date, we can rely on `remove_event` method in `CalendarModel` to remove instance of the event. Each recurrence instance has its own _recurrence identification_ that `ECal.Client` can use it to remove. After removing an instance, it can be seen from editing panel that `ECal` adds this date as exception date.

### Change log

- `remove_event` method updated in `CalendarModel` to handle removing occurrences of events
-  Fixes an issue that clicking _Remove Event_ from menu calls `remove_event` method twice, `remove_item` menu item handler connected twice later one is removed
- `add_exception` method updated in `EventMenu`, modification of exception dates is removed and instead of `update_event` method, `remove_event` method is called
- `libecal-1.2` VAPI for `get_dtstart` is corrected according to the library's documentation 

### Recommendations

- VAPI for `libecal-1.2` should be investigated deeply for similar inconsistencies between C library and Vala
- Method name for `add_exception` can be updated to `remove_occurrence`

### Revealed New Issues
- After removing occurance, UI shows events inconsistently. Scrolling will update the view and show events correctly. (This issue may appear after fixing the related issue, so new issue report may be added in future after fix is applied)

After removing occurrence of the event (wrong display)
After removing occurrence

After scrolling (correct display)
After scrolling

### DIFF
--- a/core/Model/CalendarModel.vala
+++ b/core/Model/CalendarModel.vala
@@ -178,7 +178,14 @@ public class Maya.Model.CalendarModel : Object {
     public void remove_event (E.Source source, ECal.Component event, ECal.ObjModType mod_type) {
         unowned ICal.Component comp = event.get_icalcomponent ();
         string uid = comp.get_uid ();
-        string? rid = event.has_recurrences () ? null : event.get_recurid_as_string ();
+        string? rid = null;
+
+        if (event.has_recurrences()
+                && mod_type != ECal.ObjModType.ALL) {
+            rid = event.get_recurid_as_string();
+            debug (@"Removing recurrent event '$rid'");
+        }
+
         debug (@"Removing event '$uid'");
         ECal.Client client;
         lock (source_client) {
--- a/core/Model/CalendarModel.vala
+++ b/core/Model/CalendarModel.vala
@@ -180,9 +180,9 @@ public class Maya.Model.CalendarModel : Object {
         string uid = comp.get_uid ();
         string? rid = null;
 
-        if (event.has_recurrences()
+        if (event.has_recurrences ()
                 && mod_type != ECal.ObjModType.ALL) {
-            rid = event.get_recurid_as_string();
+            rid = event.get_recurid_as_string ();
             debug (@"Removing recurrent event '$rid'");
         }
 
--- a/core/Model/CalendarModel.vala
+++ b/core/Model/CalendarModel.vala
@@ -180,8 +180,7 @@ public class Maya.Model.CalendarModel : Object {
         string uid = comp.get_uid ();
         string? rid = null;
 
-        if (event.has_recurrences ()
-                && mod_type != ECal.ObjModType.ALL) {
+        if (event.has_recurrences () && mod_type != ECal.ObjModType.ALL) {
             rid = event.get_recurid_as_string ();
             debug (@"Removing recurrent event '$rid'");
         }
--- a/vapi/libecal-1.2.vapi
+++ b/vapi/libecal-1.2.vapi
@@ -141,7 +141,7 @@ namespace ECal {
 		public void get_description_list (out GLib.SList<ECal.ComponentText> text_list);
 		public void get_dtend (out ECal.ComponentDateTime? dt);
 		public void get_dtstamp (out ICal.Time? t);
-		public void get_dtstart (out ECal.ComponentDateTime? dt);
+		public void get_dtstart (out ECal.ComponentDateTime dt);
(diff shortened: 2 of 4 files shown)


core/Model/CalendarModel.vala
```diff
@@ -178,7 +178,14 @@ public class Maya.Model.CalendarModel : Object {
     public void remove_event (E.Source source, ECal.Component event, ECal.ObjModType mod_type) {
         unowned ICal.Component comp = event.get_icalcomponent ();
         string uid = comp.get_uid ();
-        string? rid = event.has_recurrences () ? null : event.get_recurid_as_string ();
+        string? rid = null;
+
+        if (event.has_recurrences ()
+                && mod_type != ECal.ObjModType.ALL) {
+            rid = event.get_recurid_as_string ();
```
jeremypw, line 185: I think the correct elementary format for this is (line is not long enough to warrant splitting but if it is split should be after the operator):
```
if (event.has_recurrences () && mod_type != Ecal.ObjModType.ALL) {
    rid = event.get_recurid_as_string ();
    ....
}
```

vapi/libecal-1.2.vapi
```diff
@@ -141,7 +141,7 @@ namespace ECal {
 		public void get_description_list (out GLib.SList<ECal.ComponentText> text_list);
 		public void get_dtend (out ECal.ComponentDateTime? dt);
 		public void get_dtstamp (out ICal.Time? t);
-		public void get_dtstart (out ECal.ComponentDateTime? dt);
+		public void get_dtstart (out ECal.ComponentDateTime dt);
```
jeremypw, line 144: This change does not appear to be essential for the fix.

vapi/libecal-1.2.vapi
```diff
@@ -307,7 +307,7 @@ namespace ECal {
 		public weak string cn;
 		public weak string language;
 	}
-	[CCode (cheader_filename = "libecal/libecal.h", free_function = "e_cal_component_free_datetime")]
+	[CCode (cheader_filename = "libecal/libecal.h", free_function = "e_cal_component_free_datetime", has_copy_function = false, has_destroy_function = false)]
```
jeremypw, line 310: See above

vapi/libecal-1.2.vapi
mkykadir: This solves the mentioned VAPI problem in the description of the requests. It seems unecessary for resolving this issue since this request also removes "update_event" call and uses "remove_event" call. I can discard this change for this pull request but a VAPI change may be required in the future.

vapi/libecal-1.2.vapi
mkykadir: Vala compiler tries to call copy and destroy functions for this structure and it doesn't have those. This pull request doesn't need those changes anymore but this commit fixes the incorrect VAPI for this structure. I can discard this change for this pull request but a VAPI change may be required in the future.

PR #507, opened 2020-04-11
Monthly repeats by the same day should use 'by_month_day' attribute

### LINKED ISSUE #487: Repeat option does not work for "monthly"
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Describe the bug
Repeating an event daily, weekly or annually works, but not for monthly.

### To Reproduce
1. Create new event
2. Add title
3. Go to the Repeat tab
4. Repeat monthly

### Expected behavior
Should repeat events monthly.

## Platform Information
image

Please check what applies:
- [ ] I'm using the latest version from git that I've manually compiled

## Additional context
Using Google Calendar.

Currently, I use GNOME Calendar to create Monthly recurring events.

mkykadir:
Fixes #487

Monthly repeat of an event by the same day was using `by_month` attribute instead of `by_month_day` attribute in `ICal.Recurrence`. Using `by_month_day` attribute solves the problem.

### Change log

- Monthly repeats by day are now using `by_month_day` attribute
- New methods were added to ICal's VAPI to work on `by_month_day` attribute of the `ICal.Recurrence` according to the documentation of the library

### DIFF
--- a/src/EventEdition/RepeatPanel.vala
+++ b/src/EventEdition/RepeatPanel.vala
@@ -17,7 +17,7 @@
  * Authored by: Jaap Broekhuizen
  */
 
- public class Maya.View.EventEdition.RepeatPanel : Gtk.Grid {
+public class Maya.View.EventEdition.RepeatPanel : Gtk.Grid {
     private EventDialog parent_dialog;
     private Gtk.Switch repeat_switch;
     private Gtk.ComboBoxText repeat_combobox;
--- a/src/EventEdition/RepeatPanel.vala
+++ b/src/EventEdition/RepeatPanel.vala
@@ -17,7 +17,7 @@
  * Authored by: Jaap Broekhuizen
  */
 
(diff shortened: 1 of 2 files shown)


src/EventEdition/RepeatPanel.vala
```diff
@@ -17,7 +17,7 @@
  * Authored by: Jaap Broekhuizen
  */
 
-public class Maya.View.EventEdition.RepeatPanel : Gtk.Grid {
+ public class Maya.View.EventEdition.RepeatPanel : Gtk.Grid {
```
tintou, line 20:
```suggestion
public class Maya.View.EventEdition.RepeatPanel : Gtk.Grid {
```
looks like an oops :slightly_smiling_face:

mkykadir: I don't know how this appeared :) Thanks!

Issue #19, opened 2017-04-17
Offer various 'zoom' levels


elementaryBot:
Like Day, Week, Month and Year

Launchpad Details: [#LP1073336](https://bugs.launchpad.net/bugs/1073336) Alexander Wilms - 2012-10-30 21:28:27 +0000

elementaryBot:
After reading the interview with Maxwell Barian ( http://elementaryos.org/journal/interview-maya-developer-maxwell-barvian ) this feature seems to be unwanted in the early maya releases.

Should this bug be fixed cause of a new design decision? 
Adding more views would break the concept of the date chooser in the headerbar.   

Launchpad Details: [#LPC](https://bugs.launchpad.net/maya/+bug/1073336/comments/1) Steffen Schuhmann - 2014-07-03 16:42:42 +0000

elementaryBot:
While redesigning the interface this bug should also be included

https://bugs.launchpad.net/maya/+bug/1222382 

Launchpad Details: [#LPC](https://bugs.launchpad.net/maya/+bug/1073336/comments/2) Steffen Schuhmann - 2014-07-17 07:43:36 +0000

elementaryBot:
4.5 years later now. Any progress? This is a very useful feature!

Launchpad Details: [#LPC](https://bugs.launchpad.net/maya/+bug/1073336/comments/3) Kees de Kooter - 2017-02-26 23:20:54 +0000

marbetschar:
Is this still considered to be done eventually ...?

Coming from macOS Calendar I'm missing the weekly view. For me the weekly view just provides the right balance between overview and detail. I nearly never used the monthly view at all, since I can't actually see what's going on several days in a row.

Wondering if others would also love to get a weekly "zoom level"...?

ni-ka:
@marbetschar I confirm it is highly needed, the month view becomes really hard to use if you have many calendars with recurring meetings.

danrabbit:
I’m going to close this as a duplicate of #500 since there seems to be more discussion there

PR #513, opened 2020-04-21
WeekView

### LINKED ISSUE #500: Add day, week and year view mode
## Prerequisites
- [ ] I have searched open and closed issues for duplicates.

## Feature
The Calendar is such a good application and I understand the minimalist philosophy behind it but when we use it there's no options for different views such as: day, week or year. If I have a lot of events in a day, for example, I only can see them in a month view, so I can't access my daily schedule. Once again, I understand the minimalist philosophy behind it, but I think to add this feature could help users in their daily tasks. And furthermore, I know we have already the column on the right side that allows a view of my daily schedule and my week, but I think it isn't enough for some users in terms of productivity and accessibility. It'd be nicer on a entire new view mode like Gnome Calendar does. 

gnome-calendar-on-ubuntu-17

marbetschar:
**Work in Progress**

Started to work on a week view which should look similar to what macOS Calendar and GNOME Calendar provide:

macOS Calendar Week View

This Pull Request will eventually fix #500 and #19 - or at least parts of them.

### DIFF
--- a/src/WeekView/WeekGrid.vala
+++ b/src/WeekView/WeekGrid.vala
@@ -1,0 +1,43 @@
+/*-
+ * Copyright (c) 2020 elementary, Inc. (https://elementary.io)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


marbetschar:
Have to give up on this. There is some serious work effort needed to get this done which I'm not able to provide as a doing-it-in-my-spare-time volunteer.

My solution is to use GNOME Calendar for the time being.

madkoding:
> Have to give up on this. There is some serious work effort needed to get this done which I'm not able to provide as a doing-it-in-my-spare-time volunteer.
> 
> My solution is to use GNOME Calendar for the time being.

Oh please come back, this is a so good feature for calendar, and Elementary OS6 is so much mac like now, this would be awesome to have in it @marbetschar

PR #514, opened 2020-04-24
Prefer dark

danrabbit:


4jNsY6fCVqZv:
It looks like this patch is sometimes applied in Application.vala and sometimes in MainWindow.vala.
Also the positions within the files vary a lot.
Is there an ideal solution you would like to recommend especially to third party app developers?

cassidyjames (approved):
Lookin' good. :sunglasses: :shipit:

Issue #455, opened 2019-11-14
ICS files stay in UTC


PfTourneslash:
When importing an ics calendar in the app, the app doesnt convert the UTC timestamp to the computer hour (for example, I'm in France, every events start an hour early on my computer). If I'm not wrong the ICS file has to be in UTC and the soft should adapt to the computer timezone.

rbozan:
This makes Facebooks iCal for events show the wrong times

mblarsen:
I missed two meetings last week because of this :joy:

aldavigdis:
A related issue, probably not worthy of a separate ticket is that when added, events are displayed in their local time. After opening and closing the Calendar app, the times are displayed in UTC.

I had a meeting invitation from a US time zone at 9 am, which turned out to be 16:00 CET and 14:00 UTC.

mcclurgm:
Can any of you confirm whether this has been fixed? I think it should be solved now from #515, but want to confirm. It should certainly fix @aldavigdis issue, but want to be sure about the import bug.

aldavigdis:
@mcclurgm If you want me to confirm, then you can send me an invitation to my github handle at my github handle dot is.

PR #515, opened 2020-04-29
Properly handle libical timezones

### LINKED ISSUE #92: Timezones are not handled correctly
I have connected my Calendar application to two of my Google Calendars.

Events show fine on the widget (eg. if the event is today). But in the Calendar application, they show fine on the calendar itself, but not on the selected day preview.

Also it shows the time in the timezone where the Event came from, I think you should change that to the timezone where I am. On screenshot 2 it shows 12:00-12:30 while the actual time here will be 19:00-19:30.

Screenshots:
http://s15.postimg.org/z9t259ipn/Calendar1.jpg
http://s12.postimg.org/bhjp6d47x/Calendar2.jpg

Launchpad Details: [#LP1626235](https://bugs.launchpad.net/bugs/1626235) Dimitar - 2016-09-21 19:17:39 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44229500-timezones-are-not-handled-correctly?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #92, and possibly others like #455. 

This PR fixes the handling of libical timezones in `Utils.timezone_from_ical` and `Utils.get_local_datetimes_from_icalcomponent`. The previous method, using just `ICal.Time.get_timezone` failed to capture any of the timezone information on my system, like in #92. This fixes it, based on the implementation of GNOME Calendar (from [gcal-event.c](https://gitlab.gnome.org/GNOME/gnome-calendar/-/blob/master/src/core/gcal-event.c)). It uses the TZID data as well, which is included in more events.

It also fixes the timezone behavior in `Utils.get_local_datetimes_from_icalcomponent`. Previously, this method did not convert events to the local timezone, leaving them in the source timezone. This fixes that behavior.

Currently, I've just added the infrastructure to handle TZIDs and made no other changes. I need to add more handling of other cases. But this should establish the infrastructure of how we need to fix this issue.

It's still a work in progress in many ways. Here's what I'm still working on. If anyone with more Vala experience has some input, I'd really appreciate it!
- [x] I have not figured out how `ICal.Time.get_timezone` works in Vala. In GNOME Calendar, it returns NULL most of the time (see line 180 in their code). However, when I call it from Vala, it returns a timezone with offset +0. This means that I haven't actually figured out how to use `get_timezone`, which I expect is a critical piece of getting the whole infrastructure to work.
- [X] I have not checked events with non-libical TZIDs, since I don't have any. 
- [ ] I also haven't checked floating events.
- [ ] I needed to change the vapi for `icaltimezone_get_utc_offset` (and similarly `icaltimezone_get_utc_offset_of_utc_time`, although I don't use it in the code). The auto-generated one mistakenly called both parameters outputs, when the first is actually input. [(Valadoc)](https://valadoc.org/libical-glib/ICal.Timezone.get_utc_offset.html) I'm not sure if editing the vapi file in this way is allowed (its header comment suggests not).


mcclurgm:
I just added some code to fix some of the issues above, although maybe not well. Again it's based on GNOME Calendar, and this time I updated it to more closely follow the 3.36 codebase instead of 3.28 that I was working from before. 

It handles cases where the events have no timezone data using a default, which just returns the UTC timezone. It should handle all-day events more elegantly (I don't believe they ever include timezone info, so they should be handled null by default, but this may be worth being explicit about). 
It also actually uses `ICal.Time.get_timezone`, although I'm still very worried about that. To my knowledge it will still give UTC offset of 0 when it should return null. This may not be an issue, because that's the same result as if it was null and went to the else case, but I'm concerned about this unexpected behavior.

mcclurgm:
OK, I'm feeling pretty confident in this PR now, so I'm marking it ready for review. There are a few outstanding questions, but to my knowledge this PR fixes the issue completely in its current state.
So if they don't matter, then I believe this is good to go.

These are the issues. I would appreciate some help testing them, if possible:
- [ ] I can't figure out how to get floating events into libical, so I haven't been able to test that.
- [ ] GNOME Calendar has a lot more infrastructure than we have. They allow `hour_string` to include seconds, not just hours and minutes. They also have a check that `offset` (their name for `interval`) is actually in seconds, not milliseconds. We don't account for either of these. I can't come up with any reasons why we'd need them, but I imagine they exist for some reason.

The last one is more fundamental. In my testing, I think I've figured out the issue with `get_utc_offset` that was causing the problem in the first place. So I believe that the issues inside of `timezone_from_ical` could maybe be fixed by a one line change: replacing
```vala
var interval = date.get_timezone ().get_utc_offset (null, out is_daylight);
```
with
```vala
var interval = date.get_timezone ().get_utc_offset (date, out is_daylight);
```
This would mean that all of the cases I added to the top of the function may not be necessary. I'm a little hesitant to get rid of them, since it gives us more robust error handling at any rate. It may be overkill though, and I can see an argument that for simplicity, we should get rid of it all and do the simple fix.

---

 Just for completeness, here's what I've tested so far:
- Events from a variety of timezones are working, including ones with 30- or 45-minute offsets. 
- Invalid `tzid` are handled properly now (line 139). 
- All my all-day events are working as expected. 
- Editing an event works for the most part, but loses its timezone information. (I don't consider this a blocking problem, since the actual time of the event is preserved, it just changes its timezone to the local timezone on save).

Also, since I fixed the call to `get_utc_offset` to include `time`, I believe that `ICal.Time.get_timezone` is now working properly. I haven't seen a case where it's called and returns false information as before. I believe the issue was with passing it a `null` time.

core/Utils.vala
```diff
@@ -76,11 +76,49 @@ namespace Maya.Util {
...
 
     /**
-     * Converts the given TimeType to a DateTime.
+     * Gets the timezone of the given TimeType as a GLib.TimeZone.
      */
     private TimeZone timezone_from_ical (ICal.Time date) {
+        // Special case: return default UTC time zone for all-day events
+        if (date.is_date ()) {
+            debug ("Given date is 'DATE' type, not 'DATE_TIME': Using timezone UTC");
+            return new TimeZone.utc ();
```
jeremypw, line 85: This form of constructor does not appear in libical.vapi in the source code on on https://valadoc.org/libical-glib/ICal.Timezone.html yet the code compiles so I am not sure what is happening here.

core/Utils.vala
```diff
@@ -76,11 +76,49 @@ namespace Maya.Util {
     }
 
     /**
-     * Converts the given TimeType to a DateTime.
+     * Gets the timezone of the given TimeType as a GLib.TimeZone.
      */
     private TimeZone timezone_from_ical (ICal.Time date) {
```
jeremypw, line 81: To avoid confusion with ICal.TimeZone better to explicitly state GLib.TimeZone here.

mcclurgm: Good point. I just pushed a new commit to make this explicit.

jeremypw:
Need to update next release appdata (v5.0.5).  Also it would be nice to add some unit tests for the timezone_from_ical function.  Other than that this is looking good!

mcclurgm:
Sure thing. Can I ask for a little help with these though?

Since the version of appdata on master doesn't have 5.0.5 yet, should I manually add that section, or is there something I should merge into my branch?

Is there a way to generate tests for `timezone_from_ical`, which is a private function? My impression is that I create the tests in a secondary `Tests` directory in `core`, which calls functions from the library. (I'm following the model from the Music app). But since it's private, I can't access it from the test file. Should I add a public function to `Utils.vala` that adds the tests there, and then I call that function from the Tests directory, or something like that?

Thanks. Sorry for adding more work for you to review this, but this setup is pretty new to me.

jeremypw:
@mcclurgm Yes, you need to add a new release section. I just copy the previous one, delete the `<li><\li>` lines (except "Updated translations") , bump the version a minor point and change the date the current date. The add a line for this PR. e.g.:
```
    <release version="5.0.5" date="2020-05-20" urgency="medium">
      <description>
      <p>Minor Updates:</p>
      <ul>
        <li>Correctly handle timezones</li>
        <li>Updated translations</li>
      </ul>
      </description>
    </release>
```

The final form of the release data will be set when the release is made.  If another PR gets merged first you may have to fix a conflict so its best to make this the last commit if possible.

jeremypw:
The test would have to use the public function that uses `timezone_from_ical` if that can provide enough coverage otherwise there is no particular reason why `timezone_from_ical` needs to be private as it is in Utils.

mcclurgm:
Great, thank you! I'll get on it.

mcclurgm:
Sorry it's been a while. I think this is ready to go now!

The big functional change is I changed the TZID parsing to use GLib's builtin method. This gets all the timezone info (including daylight savings changes, etc), which may be useful down the line. I figure it can't hurt.

I just added some unit tests, which I cover all of the "coverting time properly" cases I could think of. I couldn't figure out how to test the actual TZID parsing functionality. Basically, I don't know how to get a custom TZID (or lack of TZID) into libical, so my tests only ever use TZID's that are known to work.

One last weird thing. I had to do a bit of tag with the compiler and linter, and it left me a little confused. On my machine, valac is giving me warnings about using the `new` keyword for the ICal.Time struct. But on the CI server, the build failed without it. I imagine I must be using a different version of Vala or something?

jeremypw:
@mcclurgm `libical.vapi` defines Ical.TIme as a `struct` but according to valadoc.org it is a GObject class.  Bionic has libecal1.2-19 which depends on libical3 but `meson.build` specifies that depends on libical2 so that appears to be incorrect.  We may be able to drop the internal libical2 vapi. I'll look into it a bit more.

I think CI is done with Focal libraries now (libical3, libecal-2) hence requiring the `new` form of construction.

jeremypw:
The `libical-glib.vapi` file shipped with the Bionic `libical-dev` package contains the same error as noted above in the `get_utc_offset` functions so there seems little choice but to include a patched version of `libical-glib.vapi` in this project for now.

jeremypw:
I had a go at trying to sort out the library mess but it proved to be more trouble than its worth given we are transitioning to Focal now.  At least this builds and passes CI now, not withstanding the warnings and the .vapi patch.

vapi/libical.vapi
```diff
@@ -1464,9 +1464,9 @@ namespace ICal {
 		[CCode (cname = "icaltimezone_get_tznames")]
 		public unowned string get_tznames ();
 		[CCode (cname = "icaltimezone_get_utc_offset")]
-		public int get_utc_offset (out ICal.Time? tt, out int is_daylight);
+		public int get_utc_offset (ICal.Time? tt, out int is_daylight);
 		[CCode (cname = "icaltimezone_get_utc_offset_of_utc_time")]
-		public int get_utc_offset_of_utc_time (out ICal.Time tt, out int is_daylight);
+		public int get_utc_offset_of_utc_time (ICal.Time tt, out int is_daylight);
```
jeremypw, line 1469: Probably should add a comment that this has been manually patched  (lines 1467 & 1469)

mcclurgm:
Good point on the vapi comment, I just fixed that. 

As for the bad vapi, my assumption is this was probably an issue with the gtk-doc or introspection data for the older version of libical-glib. If Focal updated its version, then that should hopefully fix it, but we'll see. For the moment, I agree this is probably the best we can do. If it was a real `GLib.Object` that would be amazing, because it would let me unit test things like the TZID from my last comment by making mock subclasses. But for the moment it seems like we're kind of stuck with it being a struct?

jeremypw:
As far as I can tell it *is* an object because libical3 is installed - that is why you have to use the new keyword to pass CI - it is just that the .vapi tells the compiler it is a struct in Hera.  I tried a bit to hack around it but it just got too messy.  I asked @tintou to have a look at this when he has time to get the view of a more experienced dev.

mcclurgm:
Alright, thanks for looking into that. If it's working with the current vapi, I don't think it's too big an issue, especially since Hera is still on libical 1.2 (although that probably changes for the Focal update). I'll defer to someone who knows more than me though!

PR #516, opened 2020-05-08
New calmodel method to replace event safely; use for events with rules

jeremypw:
Allows editing a non-recurring event to make it a recurring event and vice versa.

There are some restrictions that the UI does not currently prevent:

The recurrence must be the same day as the original event.
The original event must not span more than one day.

During development some orphaned objects and undeletable events were created.  Although the risk of this is now reduced it is recommended a separate calendar is used for testing.

TO FIX:
* Editing a repeating event corrupts it - a spurious Saturday repeat is added for unknown reasons and recurrences disappear - and cannot be added back.  The event has to be deleted and readded before repeats can be added again.


jeremypw:
There is an outstanding issue that when editing an event recurring weekly, the Repeat panel has _2_ days selected when opened (the correct day _and_ Saturday).  This does not matter if all recurrences are removed but it is recommended not to try to edit the recurrences in any other way at present.

jeremypw:
It appears that when a recurring event is edited to remove the recurrence (turn it off) the recurrence is not actually removed. `remove_object` with a non-null `rid` and ObjModType.ALL removes all recurring events but not the `rid` from the event.  Maybe the only way to do that is to delete the entire event.

jeremypw:
A better solution has been found.

PR #520, opened 2020-05-13
Duplicate event

jeremypw:
Based on #519 which should be reviewed first.

The main additional code is in commit https://github.com/elementary/calendar/commit/f0985d09532d1d79b4c4c5eac6be24f3278ad8e2

The existing event is cloned and the uid mangled (cloning does not assign a new uid) so that it is distinct from the original event.  This is then given to the EventDialog for modification.  When the updating the model fails (because the event has not yet been added) the model adds it instead.

This works also with recurring events and it is not necessary to change any of the details in the event (although normally it would be expected to do so).


fabiozaramella:
Thanks for working on this! Could you solve the conflicts please? 😊

jeremypw:
I seem to be getting an issue where the context menu is empty after merging - but that is occurring with master too.

jeremypw:
@fabiozaramella Should be OK now.

fabiozaramella (approved):
Seems to work as expected 👍🏼️🚀️

PR #522, opened 2020-05-25
Use new accent color variables

lainsce:
Restores functionality by using newer color vars.

Screenshot from 2020-05-25 01 49 06

Requires usage of new stylesheet.


data/style/Grid.css
```diff
@@ -18,13 +18,13 @@
 .cell {
     background-color: @base_color;
     border-radius: 0;
-    border-top: solid 1px @menu_separator;
-    border-left: solid 1px @menu_separator;
+    border-top: solid 1px alpha (black, 0.15);
```
danrabbit, line 21: I think I would prefer to add `menu_separator` to Greenfield so we can remain consistent

lainsce: Alright, reverting that.

danrabbit (approved):

PR #523, opened 2020-05-25
Make Add New Calendar button centered to look better

lainsce:
Old:
Screenshot from 2020-05-25 18 34 29

New:
Screenshot from 2020-05-25 18 35 00

Bonus: One less Label object


lainsce:
True. Closing…

PR #524, opened 2020-05-25
SourceSelector: Use Gtk.ModelButton instead of style class

danrabbit:
Supersedes #523


danrabbit:
@jeremypw should be fixed! Thanks for testing :)

Issue #480, opened 2019-12-29
"Today" indicator did not update overnight


panicsteve:
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Describe the bug
I put my laptop to sleep on the evening of Dec 28, and woke it up on the morning of Dec 29.

Pressing the "Today" button correctly takes me to Dec 29, but the green circle indicating "Today" is still on Dec 28.  (See screenshot)

### To Reproduce
1. Put computer to sleep
2. Wait for date to change
3. Wake computer

### Expected behavior
Green indicator of "Today" should be the same date that the clicking the "Today" button takes you to.

### Screenshots or screen recordings
![Screenshot from 2019-12-29 08 47 50@2x](https://user-images.githubusercontent.com/196190/71559885-409f8800-2a18-11ea-80ac-89772d9718af.png)

### Logs
Can provide logs if necessary. Please let me know where to find them.

## Platform Information
![Screenshot from 2019-12-29 08 51 19@2x](https://user-images.githubusercontent.com/196190/71559927-688eeb80-2a18-11ea-92e0-11a60b29e82f.png)

- [ ] I'm using the latest version from git that I've manually compiled
- [X] I'm using the latest released stable version

## Additional context
N/A

4jNsY6fCVqZv:
Can probably be solved in the context of this issue:
https://github.com/elementary/calendar/issues/34
The question for me is: If I leave Calendar app open overnight, then wake up the device from hibernation the following day, should Calendar automatically adjust the current day to the network or device date?

mcclurgm:
This is a duplicate of #260. It has more info though, so maybe this one should stay open?

4jNsY6fCVqZv:
@mcclurgm 

> These are two different things, although once the green dot behavior is fixed then maybe we can consider whether to change the gray dot behavior too.

means you could see yourself working on a solution for https://github.com/elementary/calendar/issues/34 as well?

4jNsY6fCVqZv:
@mcclurgm 

> These are two different things, although once the green dot behavior is fixed then maybe we can consider whether to change the gray dot behavior too.

means that you can imagine working on a solution for https://github.com/elementary/calendar/issues/34 as well?

mcclurgm:
I'm not sure honestly. I think that's a design issue more than an implementation issue for the moment, but if the core team agrees to do that then I'd be happy to give it a try.

PR #531, opened 2020-06-11
Remove duplicate grid construction in CalendarView

mcclurgm:
TLDR On its construction, `CalendarView` creates a duplicate object that can be erroneously seen by other parts of the code. I removed this.

The issue is that `CalendarView.vala` creates some `Grid` objects that it doesn't need, which can cause some odd behavior. When a `CalendarView` is constructed, it first explicitly calls `create_big_grid` in the `construct` method, line 59. This grid never gets populated with a model, as far as I can tell. It calls `create_big_grid` again in the `sync_with_model` function, which as far as I can tell simply creates a new `Maya.Grid`/`Gtk.Grid` combo that displaces the old one. If the user changes months, then both of the existing grids will be destroyed by the lambda in line 82. But if not, the old one sticks around and can receive and dispatch signals. This created some odd behavior I noticed in a comment to #480 ([link](https://github.com/elementary/calendar/issues/480#issuecomment-642200274)), where I was seeing a callback twice, when I was expecting once. One of these callbacks was being captured by the old `Grid`, which didn't have a model associated with it and errored out of what it should have done.

In this PR, I simply removed the first call to `create_big_grid` and depended on `sync_with_model` to create the visible grid. To my knowledge, this hasn't broken anything, although I'm not quite sure where to test it that could break.

I also took the liberty of renaming the `Maya.Grid` into something more distinguishable from the `Gtk.Grid`, because it took me a while to realize which was which. I hope that's OK to include in this PR.


mcclurgm:
There are no public calls to `make_big_grid`, so I can make it private no problem. I also removed `big_grid` as a member variable, since it's also never used anywhere else. So those changes both work fine. 

I haven't tried to merge the two methods yet. To me there's a pretty clear distinction between what they do (create a set of widgets vs populate them with data), but member variables like `header` and `week_labels` are only necessary because they're split up and could otherwise be removed. I could be convinced either way I suppose... It should be easy to do though, since nothing outside of `CalendarView.vala` uses either of them.

jeremypw:
From other code clean-up PRs I've seen from elementary devs, it seems to be in favour to merge functions that are called only once into the calling function or closure.  If you need to clarify the purpose you can use a comment.  However, I also have use single-call functions in the past to break the code up into smaller, more understandable pieces so I can see both sides.  As it is not essential for this PR leave that for now.

jeremypw:
This introduces two new critical warnings in the terminal on startup:
```
** (io.elementary.calendar:7134): CRITICAL **: 18:50:42.684: maya_view_grid_get_grid_range: assertion 'self != NULL' failed

** (io.elementary.calendar:7134): CRITICAL **: 18:50:42.684: maya_view_grid_get_grid_range: assertion 'self != NULL' failed

```

While these do not have any fatal effect it would be as well to fix them.

mcclurgm:
Alright, here's a quick and dirty fix. It gets the job done, but if I wanted to give it some more effort I could do a bit more surgery on the code to make it more cohesive. 

If it's really true that `days_grid.grid_range` is only null when `days_grid` is (which I think must be, since it's always set in `sync_with_model`, which is where `days_grid` is initialized), then I could simplify some more. I could combine the check on line 220 with the big if statement at the start then, which would make a small block at the beginning to deal with pre-existing grids, and then the rest would be universal. I have to do some more looking to see if this will work though.

mcclurgm:
Alright, I'm pretty confident in this new change. It's a little riskier, since it assumes that `days_grid` will always have a nonnull `grid_range` once it's initialized. This seems to be true in my analysis. It's initialized in `sync_with_model`, which also gives it a (non-nullable) range. And any further calls to `set_range` will set a nonnull range too. So as far as I can tell, after the very first call to `sync_with_model`, `grid_range` will always have a value when `days_grid` does (which is always). I've also done some basic testing, and I haven't found a situation where either would be null after the constructor finishes.

So with that, checking for `grid_range` is equivalent to checking for `days_grid`. Then I can simplify some of the `grid_range` checks together with `days_grid`, so I can condense all of the "existing grid" handling into a single block at the start. If you'd rather keep this PR simpler, I can revert it.

src/Grid/CalendarView.vala
```diff
@@ -196,17 +196,26 @@ public class Maya.View.CalendarView : Gtk.Grid {
...
-            if (days_grid.grid_range != null)
-                previous_first = days_grid.grid_range.first_dt;
+
+            // Transition in the proper direction
+            var previous_first = days_grid.grid_range.first_dt;
+            if (previous_first.compare (days_grid.grid_range.first_dt) == -1) {
+                stack.transition_type = Gtk.StackTransitionType.SLIDE_UP;
+            } else {
+                stack.transition_type = Gtk.StackTransitionType.SLIDE_DOWN;
+            }
```
jeremypw, line 212: This does not work - you are comparing two things that are the same so the comparison yields `0`.  The transition is always `SLIDE_DOWN`.

mcclurgm: 😬️ Yeah, you're right. I forgot that it's being set part way through... I'm just going to undo that commit and leave the minimal change.

jeremypw:
Can't `date_range` be made a non-null construct property of `Grid`? Grid gets recreated on every `sync_model` (in `create_big_grid`) doesnt it?

mcclurgm:
AFAIK yes, it can be. It's only created in `sync_with_model` (or really `create_big_grid`) and then `grid_range` is immediately set. It would require a bit of refactoring, and maybe then it would make sense to merge `sync_with_model` with `create_big_grid`, since the construction is done in `create_big_grid`. I think this might be a thing to leave for a different PR to keep this one simpler?

jeremypw:
OK.  The alternative would be to use a re-usable widget instead of creating a new one, but if we stick to recreating widgets we can fix their properties during construction and know they cannot change.

mcclurgm:
I agree making a reusable widget could be nice, but it would regress some things (we couldn't animate between months, since it would be the same widget). That sort of redesign is probably far above my skill level though.

jeremypw:
Ah, that true about the animation so we will keep recreating the widget.

mcclurgm:
Just fixed the merge conflict. Should be ready for review again.

PR #532, opened 2020-06-12
Keep today state updated

### LINKED ISSUE #480: "Today" indicator did not update overnight

mcclurgm:
Fixes #480.

This creates a new class to keep track of the time, and notify the app when it's time to update the current day. I plugged this signal into 2 elements, the calendar `Grid` and the sidebar `AgendaView`:
1. It updates the highlighted date on the grid without changing the selected day or month. It just removes the highlight when the current day wraps over to the next month off screen. 
2. It updates the sidebar in a fairly simple way, by simply telling it to keep its current data but re-filter it. It's possible that if someone keeps their calendar open for months, the current day could run off the end of the model (since I don't update any of that), but I don't think that's an issue because the sidebar doesn't display any events more than 2 months past the selected month.

These are all the parts of the app that I can think of that have a "today" state. If there are more though, let me know and I'll try to work with them.


src/TimeManager.vala
```diff
@@ -0,0 +1,41 @@
...
+            timeout_id = 0;
+        }
+
+        // Sets a timeout that will call at the end of the day.
+        // Creates a timer for the time until end of day, and adds an extra
+        // second to be sure it's called after midnight.
+        var now = new DateTime.now_local ();
+        var tomorrow = Util.strip_time (now.add_days (1));
+        var interval = (tomorrow.difference (now) + 1) / 1000000;
+        timeout_id = GLib.Timeout.add_seconds ((uint) interval, callback_update_today);
```
jeremypw, line 31: Does this work if the computer is suspended/hibernated?

mcclurgm: Nope, thanks for thinking of that. Working on it, and thanks for the suggestion.

mcclurgm: Ok, I added functionality to deal with this. It updates the date and sets up a new callback when the computer resumes from sleep. To the best of my knowledge, it didn't change the normal operation and sets up a proper (regular) callback when it's done.

I also have some work to update faster when settings are being changed ([like in Wingpanel too](https://github.com/elementary/wingpanel-indicator-datetime/blob/fb6281eb0302121f503e2f2fcba268e5e6294b92/src/Services/TimeManager.vala#L63)) but I think that may be another PR since it may involve some design work.

src/TimeManager.vala
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        // Sets a timeout that will call at the end of the day.
+        // Creates a timer for the time until end of day, and adds an extra
+        // second to be sure it's called after midnight.
+        var now = new DateTime.now_local ();
+        var tomorrow = Util.strip_time (now.add_days (1));
+        var interval = (tomorrow.difference (now) + 1) / 1000000;
+        assert (interval >= 0);
+        timeout_id = GLib.Timeout.add_seconds ((uint) interval, callback_update_today);
```
jeremypw, line 53: I prefer to use a closure with an explicit `return Source.REMOVE` at the end.  Looking at the `callback_update_today` function separately it is not obvious why it needs to return a boolean (and it is called elsewhere discarding the return value).

src/TimeManager.vala
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private bool callback_update_today () {
+        debug ("Updating today");
+        on_update_today ();
+        setup_today_callback ();
+        return false;
+    }
+
+    private void setup_today_callback () {
```
jeremypw, line 39: You are really setting up the timeout not the callback?

src/TimeManager.vala
mcclurgm: > I prefer to use a closure with an explicit return Source.REMOVE at the end.

I don't quite understand what you mean. Do you mean wrapping up `callback_update_today` (or whatever I'd rename it to) into a closure instead of a whole function? I broke it out like that so I could reuse it from the sleep signal, although I suppose it's such a small function that copy/pasting it could be fine. 

Were you thinking of something like this?
```vala
private void update_today () {
    debug ("Updating today");
    on_update_today ();
    setup_today_timeout ();
}

private void setup_today_timeout () {
    ...
    timeout_id = GLib.Timeout.add_seconds ((uint) interval, () => {
        update_today ();
        return GLib.Source.REMOVE;
    });
}
```

jeremypw: Yes, that it.  You can still re-use `update_today` in the `prepare_for_sleep` handler, so you are not having to repeat functional code.

jeremypw:
The function Util.strip_time () no longer exists in master so you will need to work around that.

mcclurgm:
Yeah. I think I might wait until the whole Calendar.Store thing is done, since that's a lot of big changes to keep up with in real time.

jeremypw:
OK, send me another review request when you are ready.

mcclurgm:
On a closer look, I think that's the only place my code interfaces with anything Calendar.Store would change. I fixed it, so it's ready for review again.

mcclurgm:
Yeah, I left that for another branch but I just added it back to this one. It works in my testing, and properly sets the timeout for the new end of day when the settings server closes. I was concerned that updating this frequently would use a lot of resources, but it doesn't seem to now, although granted I have a fairly fast machine.

jeremypw:
I do not think updating once a second is going to put too much of a load - and it should only happen when the time is being changed manually which should be rarely.

mcclurgm:
Yeah, I think that's right. When I tried this last week I was showing a lot of CPU usage whenever I had time settings open, but that must have been caused by something else because I don't see it now. Seems like we're good on that side of things?

jeremypw:
Yes, I was a bit worried that if Switchboard crashed (or was closed in the terminal with `<Ctrl>C` then `timedated` Calendar would remain in fast update mode. However, in this case the `name_vanished` signal seems to be emitted after a certain timeout. I guess a user leaving both Calendar and Settings open indefinitely is such a corner case we do not need to worry about it.

I also checked what happens when you change the date/time from the commandline with `sudo date +%Y%m%d -s "20200628"`.  Neither Calendar nor Wingpanel show the change immediately but they update after a similar timeout and Calendar ends up in "slow" mode, so that's OK.

There may be some corner case I have not thought of but it is unlikely to have a serious side effect (at least, not worse than the current master).

PR #537, opened 2020-06-15
Improve handling of week start

### LINKED ISSUE #536: Model `week_starts_on` doesn't handle Sunday
This is the same as elementary/wingpanel-indicator-datetime#224, since the same code is in both.

## Prerequisites

## Describe the bug
In `CalendarModel.vala`, the first day of the week is set from `Posix.NLTime.FIRST_WEEKDAY`. The way that it's implemented doesn't work with weeks that start on Sunday (for example, in the US). From what I can tell (I can't find docs on `FIRST_WEKDAY` or `nl_langinfo(_NL_TIME_FIRST_WEEKDAY)`, which I think is what this actually calls), `langinfo` uses Sunday = 1 through Saturday = 7. GLib uses Monday = 1 through Sunday = 7. So the method to subtract 1 from the `Posix` call works when the weekday is Monday or later, but Sunday will become 0. This is the enum value of `BAD_WEEKDAY`. Instead, it should wrap around to 7.

GNOME Calendar uses another call to `WEEK_DAY1` to do this "wrap around" behavior. I've tried to implement this in Vala, but I can't get that value without a segfault, so I'm not sure how to fix this. FWIW, here's their code:
```c
  union { unsigned int word; char *string; } langinfo;
  gint week_1stday = 0;
  gint first_weekday = 1;
  guint week_origin;

  langinfo.string = nl_langinfo (_NL_TIME_FIRST_WEEKDAY);
  first_weekday = langinfo.string[0];
  langinfo.string = nl_langinfo (_NL_TIME_WEEK_1STDAY);
  week_origin = langinfo.word;
  if (week_origin == 19971130) /* Sunday */
    week_1stday = 0;
  else if (week_origin == 19971201) /* Monday */
    week_1stday = 1;
  else
    g_warning ("Unknown value of _NL_TIME_WEEK_1STDAY.\n");

  week_start = (week_1stday + first_weekday - 1) % 7;
```

## Platform Information

- [x] I'm using the latest version from git that I've manually compiled
- [ ] I'm using the latest released stable version

about

mcclurgm:
Fixes #536.

First, it deals with the Sunday issue described there by explicitly converting the resulting 0 values for Sunday to 7 (which is what GLib uses).

Second, I added more support for locales based on the implementation in `Gtk.Calendar`. It requires some wrangling of POSIX standards which is all very confusing, but it should work in most cases I'm aware of. The potential issue is that some locales start their week numbering with Sunday, and some with Monday. When converting, we have to account for this. The background for the standards issues comes from a nice glib [wiki post](https://sourceware.org/glibc/wiki/Locales).


core/Model/CalendarModel.vala
```diff
@@ -294,6 +290,38 @@ public class Maya.Model.CalendarModel : Object {
...
+        // Set the "baseline" for start of week: Sunday or Monday?
+        uint week_day1 = (uint) Posix.NLTime.WEEK_1STDAY.to_string ();
+        debug (@"WEEK_1STDAY: $(week_day1)");
+        var week_1stday = 0; // Default to 0 if unrecognized data
+        if (week_day1 == 19971130) { // Sunday
+            week_1stday = 0;
+        } else if (week_day1 == 19971201) { // Monday
+            week_1stday = 1;
+        } else {
+            warning (@"Unknown value of _NL_TIME_WEEK_1STDAY: @(week_day1)");
```
danrabbit, line 308: The linter isn't happy about this. Could use printf syntax instead:
```suggestion
            warning ("Unknown value of _NL_TIME_WEEK_1STDAY: %s", week_day1);
```

core/Model/CalendarModel.vala
```diff
@@ -294,6 +290,38 @@ public class Maya.Model.CalendarModel : Object {
...
+        }
+        debug (@"$week_1stday");
```
danrabbit, line 310:
```suggestion
        debug (week_1stday);
```

core/Model/CalendarModel.vala
```diff
@@ -294,6 +290,38 @@ public class Maya.Model.CalendarModel : Object {
...
 
+    /** Set the week_starts_on property: the first day of the week.
+     *
+     * Locale handling is based on information from
+     * https://sourceware.org/glibc/wiki/Locales
+     */
+    private GLib.DateWeekday get_week_start () {
+        // Set the "baseline" for start of week: Sunday or Monday?
+        uint week_day1 = (uint) Posix.NLTime.WEEK_1STDAY.to_string ();
+        debug (@"WEEK_1STDAY: $(week_day1)");
```
danrabbit, line 301:
```suggestion
        debug ("WEEK_1STDAY: %s", week_day1);
```

core/Model/CalendarModel.vala
mcclurgm: Interesting. Thanks! Is there something different about the vala-lint we're running on CI vs the one on apt? I have it on my computer, but when I run `io.elementary.vala-lint -d core` (or `core/Model` or just about any other directory) it doesn't give me any errors. I have the latest version from apt.

mcclurgm:
I'm going to dismiss the review because I don't think it applies anymore.

As far as I can tell, the only changes you requested were fixes for the linter errors? If so, I fixed one and realized that the other 2 weren't actually supposed to make it into the commit... 🤦🏻‍♂️️

core/Model/CalendarModel.vala
```diff
@@ -294,6 +290,36 @@ public class Maya.Model.CalendarModel : Object {
...
+        if (week_day1 == 19971130) { // Sunday
+            week_1stday = 0;
+        } else if (week_day1 == 19971201) { // Monday
+            week_1stday = 1;
+        } else {
+            warning ("Unknown value of _NL_TIME_WEEK_1STDAY: %u", week_day1);
+        }
+
+        // Get the start of week
+        int week_start_posix = Posix.NLTime.FIRST_WEEKDAY.to_string ().data[0];
```
jeremypw, line 311: Again using `int.parse ()` makes more sense to me and less reliant on implementation details.

core/Model/CalendarModel.vala
```diff
@@ -294,6 +290,36 @@ public class Maya.Model.CalendarModel : Object {
...
     //--- Helper Methods ---//
 
+    /** Set the week_starts_on property: the first day of the week.
+     *
+     * Locale handling is based on information from
+     * https://sourceware.org/glibc/wiki/Locales
+     */
+    private GLib.DateWeekday get_week_start () {
+        // Set the "baseline" for start of week: Sunday or Monday?
+        uint week_day1 = (uint) Posix.NLTime.WEEK_1STDAY.to_string ();
```
jeremypw, line 300: Casting a **string** to a **uint** looks strange.  Shouldn't you use `int.parse ()`?  Or just compare strings?

core/Model/CalendarModel.vala
```diff
@@ -294,6 +290,36 @@ public class Maya.Model.CalendarModel : Object {
...
+        var week_1stday = 0; // Default to 0 if unrecognized data
```
jeremypw, line 301: You could make it clearer by making `week_1stday` an offset between Posix and GLib i.e. `-1` or `0` (and retype/rename appropriately).  Then the formula below becomes e.g.
`var week_start = week_start_posix + offset_posix;`

core/Model/CalendarModel.vala
mcclurgm: I agree, but I don't think I can do anything else. What the call to langinfo actually produces is a `  union { unsigned int word; char *string; }`, which I don't know how to deal with otherwise in Vala. (I don't know how to deal with unions in For reference, here's what GNOME Calendar does, which can interface with its C weirdness directly:
```c
  union { unsigned int word; char *string; } langinfo;
  gint week_1stday = 0;
  gint first_weekday = 1;
  guint week_origin;

  langinfo.string = nl_langinfo (_NL_TIME_FIRST_WEEKDAY);
  first_weekday = langinfo.string[0]; // My line 311: get the first byte of the string
  langinfo.string = nl_langinfo (_NL_TIME_WEEK_1STDAY);
  week_origin = langinfo.word; // My line 300: get the word.
```

When I call `to_string ()`, it produces the string `\u0001\u0002\u0001`, which afaik is just a sequence of 3 bytes. `int.parse` doesn't work on this string and just returns 0. So I don't believe that using a parser could work. The use of `.data` just gets the first byte, which is what the C implementations are doing. (Why first byte? I have no idea, just following everyone else)

Maybe there's a better way to do this. I haven't actually found any documentation for what on earth `nl_langinfo` actually returns, so it's really tough to reason about it.

core/Model/CalendarModel.vala
mcclurgm: See my comment on the other parse suggestion. This is again the only way I could get this to compile/run. I think that what WEEK_FIRSTDAY produces is just a word in the union, and the string pointer is null. (maybe?) 

When I try to actually use the string, it uses the resulting `19971130` as the pointer address to the string array, which segfaults of course. So the only way I can figure out to not have memory issues is to take the data in memory and force Vala to use it as a number instead of an address, without ever actually referencing that address. That's why I used the cast. I just copied it from the defunct [California app](https://github.com/blankon-packages/california/blob/b3dc12893c012035fcd4fc83b383cdf8933b80f4/src/calendar/calendar-system.vala#L257), I have no idea what it's really doing (and neither do they).

core/Model/CalendarModel.vala
jeremypw: OK, its obviously more complicated than it at first appears from the Vala syntax so I may have to take the view "if it works it must be OK", unless a more knowledgeable dev wants to jump in.

core/Model/CalendarModel.vala
jeremypw: See my response above.  The only problem may be that this feels like the kind of thing that may change with a newer version of Vala, but I guess we can cross that bridge when we come to it.

mcclurgm: Good point. I hope it doesn't break, but if we have to stick with this method we can deal with that when we come to it. I can add a test to make sure it doesn't break in the future. I'll also add a comment to call out the weird behavior, since as you say the syntax doesn't make clear what's going on.

jeremypw: Adding a test would be excellent.

mcclurgm:
Ok, I just added all of that. Some comments to explain what I did, and some tests. I couldn't find any locales in Ubuntu that use a different value of WEEK_1STDAY, but this should be good enough. The main test isn't running, since it's missing a settings schema. I'm not sure what to do to fix this, since it depends on the main app code to set up the gsettings. Once that's figured out, I think this is ready to go?

core/Services/Calendar/Store.vala
```diff
@@ -72,12 +72,8 @@ public class Calendar.Store : Object {
         state_settings = new GLib.Settings ("io.elementary.calendar.savedstate");
```
jeremypw, line 72: This should be "saved-state"?

jeremypw:
@marbetschar Any idea where the "Model Tests" which are failing in CI are coming from?  I cannot find them in the source code of Calendar and when I run tests locally I only get "Core Tests" - which passes.

mcclurgm:
@jeremypw That’s from me. I added that to test the hacky nl_langinfo stuff. It’s failing to build because the settings schema doesn’t exist. I think I could just change the CI build order and make the install happen before testing so the schema exists. I hope that having a permanent settings store wouldn’t make any of the tests fragile in the future, but for now nothing uses it so it doesn’t matter.

jeremypw:
@mcclurgm Ignore me - I had the wrong branch open locally!  The tests do fail locally on this branch. I'll look into it.

jeremypw:
The test were failing locally because I did not have the Arabic locale installed.  After installing that the tests pass. The failure in CI is for a different reason :-(

jeremypw:
As the tests pass locally a quick fix may be to modify Store so that it can operate in "test-mode" - i.e. does not fail if the savedstate setting is missing.

jeremypw:
I think the problem is because it looks like the tests are run after compiling Calendar but before installing it (whereas locally it has been installed before running the tests). So the setting schema has not been installed and compiled and is unavailable to GLib.  Maybe there is a way of installing and compiling the settings schema in `/core/Tests/meson.build` before the tests are run?

jeremypw:
OK, arranging for the full settings to be available during testing proved difficult so I have pushed a branch here:  https://github.com/elementary/calendar/tree/store-allow-no-settings that allows the tests to run in an environment where the settings are not available. Please use it if you wish.

mcclurgm:
Wouldn't it just work to change the order in `main.yml` so the `ninja install` comes before `ninja test`? I assumed it would, but maybe not.

At any rate, having a testing shim seems like a better idea anyway. Good call, thanks for working on that! I included one small comment on your branch.

jeremypw:
I had feeling there was a simpler solution, but I just tried changing the test-install order (see ) but CI still failed for the same reason.  Relevant output was:

```
Installing /__w/calendar/calendar/data/io.elementary.calendar.gschema.xml to out/usr/local/share/glib-2.0/schemas
Installing /__w/calendar/calendar/core/elementary-calendar.deps to out/usr/local/share/vala/vapi
Installing /__w/calendar/calendar/build/meson-private/elementary-calendar.pc to out/usr/local/lib/x86_64-linux-gnu/pkgconfig
Running custom install script '/usr/bin/meson --internal gettext install --subdir=po --localedir=share/locale --pkgname=io.elementary.calendar'
Running custom install script '/usr/bin/python3 /__w/calendar/calendar/meson/post_install.py'
ninja: Entering directory `build'
[0/1] Running all tests.
1/2 Utils Tests                             OK       0.07 s 
2/2 Model Tests                             FAIL     0.37 s (killed by signal 5 SIGTRAP)

Ok:                    1
Expected Fail:         0

--- command ---
10:50:12 /__w/calendar/calendar/build/core/Tests/io.elementary.calendar-tests-model
--- stdout ---
# random seed: R02Se6c626211aa0fd1c12208f489e362296
1..2
# Start of Model tests
# Start of Locale tests
ok 1 /Model/Locale/nl_langinfo
Bail out! GLib-GIO-FATAL-ERROR: Settings schema 'io.elementary.calendar.savedstate' is not installed
--- stderr ---

(/__w/calendar/calendar/build/core/Tests/io.elementary.calendar-tests-model:9011): GLib-GIO-ERROR **: 10:50:12.440: Settings schema 'io.elementary.calendar.savedstate' is not installed
```
So there is still some difference in the test environment - not sure what though.  Maybe the post_install script is failing to compile the schema?

mcclurgm:
That's surprising.. Then the compatibility shim type thing is probably the easiest thing to do, and I can also see a good no-settings fallback as a side benefit.

jeremypw:
I just pushed a PR to this branch from `elementary/calendar/store-allow-no-settings`.

mcclurgm:
Finally merged that in, and it's working now!. If this is ready to go then should we merge it now or wait for the next Calendar.Store to go through? I would imagine this has some conflicts. @marbetschar what do you think?

One last thing that I remembered. I added a requirement for the Arabic locale in testing, so should I document that somewhere? It messed you up before, and it does add a new package people need to install to be able to test. Since it's just for the test, should I a comment in the test file or something? (The extra locale is kind of a pain and makes CI setup take way longer, but I can't figure out another way to get a locale with week start on Friday)

marbetschar:
@mcclurgm every merged bugfix makes it way easier to pin down issues which are truly related to Calendar.Store. Merge Conflicts are kind of inevitable anyways ;)

So from my end, feel free to merge as soon as the fix passes the review.

danrabbit:
Instead of trying to work around the gsettings requirement, have we tried changing the build command to `ninja -C build install test` to make sure we're including an install in the test step?

jeremypw:
@danrabbit Just tried your suggestion, but I still get build failure.  Relevant messages were:
```
Installing /__w/calendar/calendar/build/data/io.elementary.calendar-daemon.desktop to out/usr/local/etc/xdg/autostart
Installing /__w/calendar/calendar/data/io.elementary.calendar.gschema.xml to out/usr/local/share/glib-2.0/schemas
Installing /__w/calendar/calendar/core/elementary-calendar.deps to out/usr/local/share/vala/vapi
Installing /__w/calendar/calendar/build/meson-private/elementary-calendar.pc to out/usr/local/lib/x86_64-linux-gnu/pkgconfig
Running custom install script '/usr/bin/meson --internal gettext install --subdir=po --localedir=share/locale --pkgname=io.elementary.calendar'
Running custom install script '/usr/bin/python3 /__w/calendar/calendar/meson/post_install.py'
[1/2] Running all tests.
1/2 Utils Tests                             OK       0.07 s 
2/2 Model Tests                             FAIL     0.37 s (killed by signal 5 SIGTRAP)

Ok:                    1
Expected Fail:         0
Fail:                  1
Unexpected Pass:       0
Skipped:               0
Timeout:               0


The output from the failed tests:

2/2 Model Tests                             FAIL     0.37 s (killed by signal 5 SIGTRAP)

--- command ---
16:50:16 /__w/calendar/calendar/build/core/Tests/io.elementary.calendar-tests-model
--- stdout ---
# random seed: R02Sa86a224008953ec8d3f3b44ce02fd7ec
1..2
# Start of Model tests
# Start of Locale tests
ok 1 /Model/Locale/nl_langinfo
Bail out! GLib-GIO-FATAL-ERROR: Settings schema 'io.elementary.calendar.savedstate' is not installed
--- stderr ---

(/__w/calendar/calendar/build/core/Tests/io.elementary.calendar-tests-model:8898): GLib-GIO-ERROR **: 16:50:16.923: Settings schema 'io.elementary.calendar.savedstate' is not installed
-------
Full log written to /__w/calendar/calendar/build/meson-logs/testlog.txt
FAILED: meson-test 
/usr/bin/meson test --no-rebuild --print-errorlogs
ninja: build stopped: subcommand failed.
##[error]Process completed with exit code 1.
```

There probably is a way of getting the settings available to tests in CI but I do not know what it is.

danrabbit:
Ah bummer. It was worth a shot!

danrabbit:
@jeremypw @mcclurgm what's the status on this? Is it ready to go? It looks like CI is passing

jeremypw:
@danrabbit Looks good to go now.

Issue #529, opened 2020-06-05
Add missing palette colors to color button list


4jNsY6fCVqZv:
More specifically, it's about adding the colors Mint and Bubblegum to the color selection when creating new calendars.

I am happy to work on it and have created a PR that is ready for a review.

PR #538, opened 2020-06-16
Add mint and bubblegum to the color palette

4jNsY6fCVqZv:
Fixes https://github.com/elementary/calendar/issues/529


4jNsY6fCVqZv:
Sorry, I've been playing around a bit with the GitHub web interface, which honestly confused me a bit. Can someone of the admins therefore please delete this closed PR again, so that the repository will not get cluttered.

src/SourceDialog/SourceDialog.vala
```diff
@@ -139,6 +141,11 @@ public class Maya.View.SourceDialog : Gtk.Grid {
         color_button_green_context.add_class ("green");
         color_button_green_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
jeremypw, line 143: I think you forgot to create the new color button `color_button_mint = new Gtk.RadioButton.from_widget (color_button_red);`

4jNsY6fCVqZv: So true!

4jNsY6fCVqZv:
> I think this depends on an updated stylesheet being released as the new colors @MINT_500 and @BUBBLEGUM_500 do not appear on the buttons for me. If so, this cannot be merged yet.

The last release is from the end of February.
Is https://github.com/elementary/stylesheet/blob/5.4.2/elementary/gtk-3.0/palette.css a good place to look? Looks like mint and bubblegum were not included at this point.

> It seems odd the SourceDialog code contains both style names for colors and their hex equivalent (which may or may not be the same color) but the code could do with some cleanup in another PR anyway.

In the coming two PRs, I want to make improvements in terms of consistency and compliance. What improvements did you have in mind for the SourceDialog? Then maybe I can just put that into another PR.

src/SourceDialog/SourceDialog.vala
4jNsY6fCVqZv: Oh yes, after I wrote this line a few times before, it actually slipped through my fingers. :smile:

jeremypw:
>Is https://github.com/elementary/stylesheet/blob/5.4.2/elementary/gtk-3.0/palette.css a good place to look?

Yes. The MINT and BUBBLEGUM colors were added to this repo in April with these PRs:
https://github.com/elementary/stylesheet/pull/606
https://github.com/elementary/stylesheet/pull/607

>You tested it with Hera, not Odin, right?

Yes. Last time a tried to install Odin it would not boot on my laptop.  Hopefully I will be able to get a working install soon.

>What improvements did you have in mind for the SourceDialog?

Some suggestions would be
* DRY the code for adding color buttons.  Could use a function or a private class taking the style classes and hex value as parameters.
* Look into whether it is possible to eliminate the hex values or at least assign them to constants.

4jNsY6fCVqZv:
> Yes. The MINT and BUBBLEGUM colors were added to this repo in April

So what do you think would be a good next step?
Can this PR be merged, even if it requires Odin? From what I have seen, Greenfield should have no problem with this PR. Or would a small point release for Stylesheets be useful so that the changes can take effect before an Odin release?

A solution here probably also concerns my approach to this PR https://github.com/elementary/tasks/pull/119#issuecomment-640037848

jeremypw:
I would have thought another release of the stylesheet is due, but @danrabbit is the authority on UX matters.  If this were merged it would block any Calendar releases in advance of the stylesheet release.

danrabbit:
Yeah I think we should consider this blocked until we're sure we don't want to make any more Calendar releases into Hera. Right now it looks like we're getting a lot of good issues fixed in the calendar backend, so I think we should probably hold off.

Marking as "Blocked"

4jNsY6fCVqZv:
Should I make these changes for Calendar as well? https://github.com/elementary/tasks/pull/129

fabiozaramella:
> Should I make these changes for Calendar as well? [elementary/tasks#129](https://github.com/elementary/tasks/pull/129)

I think it makes sense.

PR #546, opened 2020-06-23
Unify reminders, guests placholder styles

### LINKED ISSUE #541: Label size should be consistent
Screenshot
Screenshot

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92263827-label-size-should-be-consistent?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #541 

I decided to use standard text size and style, with the `dim_label` style. This seems to be in line with the GNOME HIG, and I haven't found anything about this on elementary HIG. I also removed an insensitive property on one, which should be redundant.


danrabbit:
Looks good! I pushed an additional commit to https://github.com/elementary/calendar/pull/559 that uses `h3` which is what we're using in the datetime indicator, and also includes the "No Exceptions" label from the reminders panel.

Gonna close this branch and merge that one. Thanks for starting this!

mcclurgm:
Great! Is the H3 style standard for elementary, and if so, should I add something to the HIG for this? As far as I know, we have no guidance on _when_ to use the header styles, or anything for placeholders.

danrabbit:
@mcclurgm yeah that would be great if you wanted to add that to the HIG. I agree that right now there isn’t clear guidance here

mcclurgm:
Just to be sure, the style is `dim_label` and `H3` right?

PR #549, opened 2020-06-24
Calendar.Store: Unifying API for Events and Tasks

marbetschar:
This PR is a follow up of the work done in wingpanel and tasks, to use one service object providing access to events and tasks using a unifyied API:

- elementary/wingpanel-indicator-datetime#225
- https://github.com/elementary/tasks/pull/123

PR #560, opened 2020-06-27
Calendar.Store: Integrated new Calendar.Store object

marbetschar:
One of multiple PR's to break [Calendar.Store](https://github.com/elementary/calendar/pull/549) down in smaller chunks to make review easier. This one contains:

- Made the Calendar app using the all new `Calendar.Store` object

**IMPORTANT:** This is the first attempt to integrate the new Calendar.Store into Calendar. That said, this PR has not yet been extensively tested and chances are high there are bugs and/or regressions we need to fix to ensure Calendar still works properly after merging.

Please report anything you find, and I'm happy to make it work as intended.

**UPDATE:** Just a quick heads up since there was a lot of discussion going on in the review process of this PR:

I guess the best approach would be to use this PR as guidance of how a unified API for Tasks and Events may look and work, but break it down into a lot micro PR's to make review easy.

Even though this PR might not get merged, we can still keep the overall discussion here, so we have a place to talk about the bigger picture of all this work.


core/Services/Calendar/Store.vala
```diff
@@ -343,145 +694,408 @@ public class Calendar.Store : Object {
...
 
-        Idle.add (() => {
-            connected (source);
-            load_source (source);
-            return false;
-        });
+        if (client == null) {
+            critical ("No client was found for source '%s'", source.dup_display_name ());
+        } else {
+            string? uid;
```
mcclurgm, line 826: You don't do anything with `uid` here. The original code has a check to make sure the uid is set right after the `#endif`:
```vala
if (uid != null) {
    comp.set_uid (uid);
}
```

core/Services/Calendar/Store.vala
```diff
@@ -1,307 +1,658 @@
...
+
+    public void source_disconnect (E.Source source) {
+        source_disconnect_async.begin (source);
+    }
+
+    public E.Source source_get_with_uid (string uid) {
+        return registry.ref_source (uid);
+    }
+
+    public string? source_get_ancestor_display_name (E.Source source) {
```
mcclurgm, line 174: This function is never used, at least in this state.

core/Services/Calendar/Store.vala
```diff
@@ -1,307 +1,658 @@
...
+                    components.add_all (source_components.get (source.dup_uid ()).get_values ().read_only_view);
+                }
+            });
         }
+        return components;
     }
 
-    public void change_month (int relative) {
-        month_start = month_start.add_months (relative);
+    //--- Privat ECal.Component Helpers ---//
```
mcclurgm, line 524:
```suggestion
    //--- Private ECal.Component Helpers ---//
```
Small typo

core/Services/Calendar/Store.vala
```diff
@@ -1,307 +1,658 @@
...
+                break;
 
+            case ECal.ClientSourceType.TASKS:
+                registry.default_task_list = source;
+                break;
+            }
+        }
+    }
+
+    public List<E.Source>? sources_list () {
```
mcclurgm, line 305: I don't think this needs to be nullable. According to valadoc, `list_sources` can't return null.

core/Services/Calendar/Store.vala
```diff
@@ -1,307 +1,658 @@
...
+    private Gee.Collection<E.Source> sources_add_transaction;
+    private Gee.Collection<E.Source> sources_connect_transaction;
 
-    HashTable<string, ECal.Client> source_client;
-    HashTable<string, ECal.ClientView> source_view;
-    HashTable<E.Source, Gee.TreeMultiMap<string, ECal.Component>> source_events;
+    internal HashTable<string, Gee.TreeMultiMap<string, ECal.Component>> source_components;
 
-    public GLib.Queue<E.Source> calendar_trash;
+    private GLib.Queue<E.Source> sources_trashed;
```
mcclurgm, line 44: Minor style: this is a property, not a function, so it should be a noun. This would also be consistent with `calendar_trash`.

```suggestion
    private GLib.Queue<E.Source> sources_trash;
```

mcclurgm:
There's a lot of reorganizing going on here, so I tried to go through it and figure out what corresponds to what. Here's a document with what I came up with: https://gist.github.com/mcclurgm/c5e9f6a65a31c43b313976cdb7b2f0f9

I did this before the new commit (`5aa567b`), so it's a bit out of date. I may try to update it, but at any rate it should still make the new changes a little easier to parse.

core/Services/Calendar/Store.vala
marbetschar: Good catch! Although I'm not quite sure a) why this is needed and b) events are still fired as expected after this is re-added.

Re-added nevertheless for now, because it was there in the original code.

core/Services/Calendar/Store.vala
marbetschar: True that. This one [is used in the Tasks app](https://github.com/elementary/tasks/blob/2226b036bc8870e7a5944a411934bbf2cea67f58/src/MainWindow.vala#L226). I guess we should decide whether we want to use `source_get_ancestor_display_name` or `source_get_location` as for the section header names.

From what I can tell, these are two different approaches and I didn't want to mix them up - so I implemented both methods to not change behaviour in the Calendar app.

core/Services/Calendar/Store.vala
marbetschar: Fixed!

core/Services/Calendar/Store.vala
marbetschar: This function _might_ be executed too early (when `registry` is not available yet), so it has nothing to return in this case. I have to admit this might happen very rarely. Are we fine with some weird exception in this case?

core/Services/Calendar/Store.vala
marbetschar: Fixed!

marbetschar:
There is an issue when modifying the same event twice when using a Nextcloud backend. The E-Tag seems to be out of date the second time. I encountered this issue earlier and its due to sending an outdated calcomponent object to EDS for updating.

But I don't seem to be able to figure out why the component in the UI is not updated after first modification. Any ideas?

marbetschar:
**UPDATE:** Fixed the error thrown when updating the same event twice, by ensuring the UI always uses the latest ECal.Component:

https://github.com/elementary/calendar/pull/560/commits/aabde4bbc128d5625802c03d69ff3de5950887fc

marbetschar:
**Issue to Fix:** If we edit a single event and set it to repeat daily, the UI does not reflect this change correctly. In order to update the UI, we need to jump forward/backwards in the month - then the daily events are shown. The same goes, if we edit a daily event and set it to single ocurrence - the UI does handle this case wrong as well. Don't know.

core/Services/Calendar/Store.vala
```diff
@@ -1,307 +1,658 @@
...
+                break;
+        }
+    }
+
+    public void component_add (E.Source source, ECal.Component component) {
+        var views = source_get_views (source);
+
+        var components = new Gee.ArrayList<ECal.Component> ((Gee.EqualDataFunc<ECal.Component>?) Calendar.Util.ecalcomponent_equal_func);  // vala-lint=line-length
+        components.add (component);
+        components_added (components.read_only_view, source, views.read_only_view);
```
mcclurgm, line 450: The way this is designed, callbacks are called twice for every new event. This is called here, then the component is added to `transactional_components`. Then, `source_component_add` adds the component to a `Source`, which then sends the `source_icalcomponents_added` signal. That captures `transactional_components` again and calls `components_modified` for the same events.

This actually raises two questions for me. 
1. What is the purpose of `components_added_transaction`? As far as I can tell, all that happens to a component at the moment is that it's added, and then immediately removed by `source_icalcomponents_added` or a catch block in `component_add`. I don't see it used anywhere in between. And `source_icalcomponents_added` already has a list of the relevant components.
2. Why do you need to check for modified components in the component_added callback at all? If things are functioning correctly on the libecal side, the function should only be called for truly new components, unless I'm missing something. I imagine this could also answer question 1, but I'm not sure what that answer is is.

core/Services/Calendar/Store.vala
mcclurgm: Aha, got it. Then it makes sense to me as is.

mcclurgm:
One minor performance regression I found. In the new version, there's a delay after I add an event where its details don't appear in the UI, but an empty box does. I doubt this is enough to stop this PR, but something to consider.
Maybe something about this version is making things async that weren't before, so the UI gets an event before it's finalized? Although with any luck this can be solved by #519 that fixed some issues with the day boxes.

marbetschar:
@mcclurgm that's totally possible, because the new Calendar.Store optimistically shows changes in the UI before EDS replies with success or error (in case of an error, Calendar.Store reverts the changes in the UI).

This way, interactions should feel more responsive, as most of the time EDS is sucessful anyways - but this behavior has not extensively tested yet.

core/Services/Calendar/Store.vala
marbetschar: @mcclurgm this is actually needed to optimistically show changes in the UI - even though they are not confirmed by the EDS backend yet:

- When the user adds a new event, Calendar.Store immediately fires components_added to let the UI display the new event
- Then, Calendar.Store sends the new event to the EDS backend
- Once EDS backend has added the event, we need to update the event in the UI again to avoid update errors of the same event due to internally changed E-Tags

Because EDS fires a "components_added" event itself, Calendar.Store needs a way to figure out if the event in question was already added in the UI before (and therefore needs to be updated) - or if the event was never rendered before and needs to be added. That's where `components_add_transaction` comes into play: It holds all events, which were added to the UI before and only needs updating.

In case you add an event on another device for example, the EDS "components_added" event will be fired too. Because `components_add_transaction` does not contain the event in question, Calendar.Store knows it needs to add the event to the UI instead of updating it.

mcclurgm:
Aha, got it. Personally, I'd argue that showing incomplete data isn't desirable. I think that's a discussion for another time, though. 

All that's by design, so good with me. You can dismiss my comment above. One thing I'd like to do though, is add some comments to explain what's going on. This wasn't obvious to me at all. I can do that if you like and submit a PR to your branch.

core/Services/Calendar/Store.vala
```diff
@@ -26,282 +26,630 @@ public class Calendar.Store : Object {
...
-            store = new Calendar.Store ();
-        return store;
+    private Store (ECal.ClientSourceType source_type) {
+        Object (source_type: source_type);
+    }
+
+    private static Calendar.Store? event_store = null;
+    private static Calendar.Store? task_store = null;
+
+    public static Calendar.Store get_event_store () {
```
danrabbit, line 81: I'm not sure this is more clear. The convention of `get_default` is used across many apps and libraries and makes it more immediately clear that this is a singleton

mcclurgm: I think the issue here is `get_event_store` vs `get_task_store`. Since there can be both, there has to be some way of distinguishing this.

marbetschar: @mcclurgm is right; actually we are dealing with two singletons here: one for tasks and one for events. However, we can change the API to something like:

- `get_default (ECal.ClientSourceType store_type)`

or:

- `get_default_event_store ()`
- `get_default_task_store ()`

if something like this is preferred. What do you think @danrabbit?

marbetschar:
> Aha, got it. Personally, I'd argue that showing incomplete data isn't desirable. I think that's a discussion for another time, though.

@mcclurgm actually the UI should not show incomplete data (at least for a simple event without recurrences). From what I'd expect, the user provided data is already available and therefore it should be possible to display it correctly in the UI. So I guess this is a bug. Can you elaborate on what exactly you mean by "showing incomplete data"...?

> All that's by design, so good with me. You can dismiss my comment above. One thing I'd like to do though, is add some comments to explain what's going on. This wasn't obvious to me at all. I can do that if you like and submit a PR to your branch.

Feel free to do so 👍️

core/Services/Calendar/Store.vala
danrabbit: Oh interesting. Sorry I'm just reading through this. Is the task store used at all in Calendar?

marbetschar: @danrabbit thats fine. The task_store is not used in the Calendar app. An example of both stores are in use in wingpanel-indicator-datetime - even though the Calendar.Store there is a bit older.

marbetschar: FWIW: we can also use an abstract super class "Calendar.Store" with two implementations: "Calendar.TaskStore", "Calendar.EventStore" which then can expose a corresponding "get_default ()".

mcclurgm:
@marbetschar I see what you're getting at better now. To me, the "incomplete information" is the empty event box that's shown in the UI before the callback completes. This may just be a bug, but here's my current case: Create a new repeating event. For a fraction of a second, you'll see an empty cell on the first day of the event. Then the rest of the recurrences appear, and it disappears. When I add something to the calendar, I expect to only see something in the UI that's complete and exactly what I entered.

core/Services/Calendar/Store.vala
mcclurgm: Just thinking out loud, if we wanted to go that direction I can see a compelling argument for making the generic Store an abstract class and embracing polymorphism. Other functions like `view_icalcomponents_added` have branches for the different types, and it would be more readable to make them implementation details of concrete subclasses instead of adding a branching if statement with another level of indentation.

marbetschar:
@mcclurgm it does add a fair amount of complexity to the code, that's true. However, in my experience it does feel so much better to work with an App which provides immediate response in its UI - even if its not 100% accurate.

On my configuration (Dell XPS 7390 from 2020, using a Nextcloud backend connected via EDS) there is a notable lag between a change and its reflection in the UI - I'd guess its something around 1 second.

That said, working with the Calendar app just feels unresponsive. It gets even worse in the Task app, if you just want to jot in some new tasks and/or mark some of them to as completed - the app "freezes" constantly. And as a user I feel like I have to wait, until the change I did is also reflected in the UI. Otherwise I fear I loose my data or corrupt it in some way. So I can see a real problem there in terms of UX: At best case it feels awkward/annoying at worst case its downright confusing.

So we really want a way to reflect changes to the user immediately in the UI - regardless of the backend we use. In my personal opinion showing "parts of information" is even better than showing nothing. This at least indicates that "something is in the works" - instead of "the App did not respond at all seems likes its freezed".

Of course it goes without saying we should aim for showing everything correctly at first attempt. But I fear - especially in terms of recurring events - we may not easily be able to do so, since the calculations are done by the EDS backend.

core/Services/Calendar/Store.vala
marbetschar: @mcclurgm yeah, the code base might really benefit from this separation. Also I guess it would make it easier to differentiate between more Tasks or Calendar specific code. So I guess I really like the idea of having such a thing like:

```
public abstract class Calendar.Store {
  ...
}

public class Calendar.EventStore: Calendar.Store {
  public static Calendar.EventStore get_default () {
    ...
  }
  ...
}

public class Calendar.TaskStore: Calendar.Store {
  public static Calendar.TaskStore get_default () {
    ...
  }
  ...
}
```

mcclurgm: That's exactly what I was thinking. If we have a nice OO language, might as well use it :nerd_face: 

I could try to take that on or help you along once this is merged. I assume it would be better to keep the design as is for now and do more implementation detail changes in a separate review.

marbetschar:
Just a quick heads up since there was a lot of discussion going on in the review process of this PR:

From this point on I guess the best approach would be to use this PR as guidance of how a unified API for Tasks and Events may look and work, but break it down into a lot micro PR's to make review easy.

Even though this PR might not get merged, we can still keep the overall discussion here, so we have a place to talk about the bigger picture of all this work.

mcclurgm:
Alright, sounds good. Hopefully this new plan isn't too much more work. :grimacing: 

For your last comment—you have me convinced about optimistically adding components. I do have one question though. Do we want to just expose the async methods like `component_add_async` in the public API or keep the synchronous wrapper functions? It would put more responsibility on the caller to use it properly, but it would keep the Store code cleaner and give the caller the option to await the result if they want. It would also make the async nature of the call clearer.

marbetschar:
@mcclurgm I don't have a strong opinion about exposing async methods vs. keep synchronous wrapper functions as of now. My gut tells me its probably better to handle all transactional logic within Calendar.Store, to make code using its API less error prone. Of course, this increases complexity of the Calendar.Store itself - but with the benefit of simplyfing each API consuming App.

But as said, thats just a feeling. I guess we'll just have to start with the micro PR's and see where we end up 😃

mcclurgm:
I think that is the right call too. It’s 3 extra lines for a simpler interface 🙂 Maybe just mention that it’s partly async in a comment.

marbetschar:
superseded by #595.

PR #562, opened 2020-06-28
Use the official color palette names consistently

4jNsY6fCVqZv:

PR #530, opened 2020-06-05
Add missing palette colors to color button list

4jNsY6fCVqZv:
Superseded by https://github.com/elementary/calendar/pull/538

Additional improvements in the context of this PR:
- Consistent use of color names based on the official color palette for existing CSS classes and variable names.
- Bringing up the question regarding the need for WCAG compliance.

PR #565, opened 2020-06-30
Make the hex values consistent…

4jNsY6fCVqZv:
…with the color constants of ColorButton.css

### DIFF
--- a/src/SourceDialog/SourceDialog.vala
+++ b/src/SourceDialog/SourceDialog.vala
@@ -24,7 +24,7 @@ public class Maya.View.SourceDialog : Gtk.Grid {
 
     private Gtk.Entry name_entry;
     private bool set_as_default = false;
-    private string hex_color = "#da3d41";
+    private string hex_color = "#ed5353";
     private Backend current_backend;
     private Gee.Collection<PlacementWidget> backend_widgets;
     private Gtk.Grid main_grid;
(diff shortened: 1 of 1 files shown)


danrabbit:
This is intentional since (for example) `#f9c440` is only 1.6:1 contrast against white and `#e6a92a` is 2.1:1

I'm inclined to close this since it would reduce contrast quite a bit

4jNsY6fCVqZv:
Currently, the color constants mentioned in ColorButton.css do not always correspond to the hex values of the corresponding places in the SourceDialog.vala.
Also, the SourceDialog.vala contains hex values that are not part of the official color palette.

My PR would like to adjust this in an intermediate step, so that I can take care of the WCAG compliance in a further step. See my comments on this in https://github.com/elementary/calendar/pull/530

So what do you think about merging this PR and then looking again specifically at the contrasts, if the use of hex values and color constants is at least consistent with the official color palette again?

danrabbit:
Legibility here is more important to usability than consistency. Any changes need to continue to pass contrast checks

4jNsY6fCVqZv:
For comparison:

red
before: `#da3d41` -> 4,4:1 WCAG A
after: `#ed5353` -> 3,5:1 WCAG A

orange
before: `#f37329` -> 2,9:1
after: `#ffa154` -> 2,0:1

yellow
before: `#e6a92a` -> 2,1:1
after: `#f9c440` -> 1,6:1

green
before: `#81c837` -> 2,0:1
after: `#68B723` -> 2,5:1

Currently the ColorButton.css shows color constants that are not reflected by the hex values in the SourceDialog.vala.
Red, yellow and green use color values that are not part of the official color palette.

If I look at the comparison, by switching to the official green we could even slightly increase the contrast and by switching to the official red we would also retain the current WCAG compliance.

What scope for improvement do you see in terms of more consistency? At the moment it is simply not coherent in the code.

4jNsY6fCVqZv:
@danrabbit Can you look at it again?

4jNsY6fCVqZv:
For me this PR is directly connected to https://github.com/elementary/calendar/pull/562. If some colors are not supposed to correspond to the elementary color palette, I don't need to use the color name of the official palette for those, do I?

PR #569, opened 2020-06-30
RDNN gresource.xml

4jNsY6fCVqZv:


meson.build
```diff
@@ -42,7 +42,7 @@ m_dep = meson.get_compiler('c').find_library('m', required : false)
 
 gresource_calendar = gnome.compile_resources(
     'gresource_calendar',
```
4jNsY6fCVqZv, line 44: Why is `gresource_calendar` used here for naming? Should I adjust that too?

4jNsY6fCVqZv, line 44: Why is `gresource_calendar` used as name at this point?
Other than that I find this term only in calendar/src/meson.build (line 65).

ryonakano: According to [the documentation](https://mesonbuild.com/Gnome-module.html), this argument represents the name of the resource which is used for a header file name generated in the `build` directory. In my understanding, it's not decided and any name is fine.

meson.build
```diff
@@ -42,7 +42,7 @@ m_dep = meson.get_compiler('c').find_library('m', required : false)
...
-    'data/maya.gresource.xml',
+    'data/io.elementary.calendar.gresource.xml',
```
ryonakano, line 45:
```suggestion
    join_paths('data', meson.project_name() + '.gresource.xml'),
```

4jNsY6fCVqZv: Thanks, that looks like a smart improvement!
What motivated you to make this proposal? And doesn't my solution work as good as yours here?

ryonakano: Your solution does work. My suggestion is just a kind of nitpicking and not a necessary thing. :sweat_smile:

1. The app name `io.elementary.calendar` that we chose for the gresource filename is the same as the project name specified in `meson.build:1`. And you can get this value by `meson.project_name()`. If we change the project name in the future or some developer who forked Calendar and make another great app based on it with another name, this way may help.

2. using `join_paths()` instead of explicating the path separator by `/` commonly used in *nix allows some challenging developer building this app against or provide a package of it for any distros/OSes that don't use `/` as the directory separator (but this may not necessary because Calendar is designed for elementary OS and don't expect it to run on Windows or macOS)

4jNsY6fCVqZv: @ryonakano Thank you for your explanation! Do you know current distros that package elementary Calendar for their desktop environment, use no `/` paths and would therefore be happy about a solution like `join_paths()`?
@danrabbit How would you like to see such a case solved for elementary OS? Especially since the solutions within the elementary default apps look quite different as well.

4jNsY6fCVqZv: @ryonakano Thank you for your explanation! Do you know current distros that package elementary Calendar for their desktop environment, use no / paths and would therefore be happy about a solution like join_paths()?
@danrabbit How would you like to see such a case solved for elementary OS? Especially since the solutions within the elementary default apps look quite different as well.

danrabbit: @4jNsY6fCVqZv In a completely practical sense, the end result should be the same, but the suggested change here is more future proof and generally being more generic is better

Another option (that I've been favoring lately) is to name the file itself as simply `gresource.xml`. Since this will be compiled into the binary and not installed, it doesn't really need to be RDNN'd

For files that are installed (like gschemas), you can also make use of the `rename()` function in meson. I've found that this makes it easier to navigate these files in Code's sidebar for example

4jNsY6fCVqZv: thanks for your feedback!

PR #581, opened 2020-08-15
Use exclusive dtend for all day events

### LINKED ISSUE #580: End times for all-day events are incorrect
## Prerequisites

## Describe the bug
In our code, we incorrectly assign the end time of all-day events. It's one whole day earlier than it should be. I assume this is so that the date of the end time corresponds to the last day that the event occurs, instead of the following day.
This leads to issues with comparisons: for example, we say that a DateRange that is in the middle of an all-day event doesn't contain the event, because the event starts and ends before the the range. This is incorrect.

### To Reproduce
Since this is so internal, hard to say exactly. Here's one not-very-specific example. Create an all-day event that lasts only one day. Print the start and end times for that event (which you get from `icalcomponent_get_local_datetimes`). You will see that the start and end times are exactly the same.

### Expected behavior

In the iCalendar spec, it seems that a single-day all-day event starts at midnight and ends at midnight the following day. So an all-day event on 11/21 should start on 11/21 at 0:00 (12:00 AM) and end on 11/22 at 0:00.

## Platform Information

- [x] I'm using the latest version from git that I've manually compiled
- [ ] I'm using the latest released stable version

## Additional context

As I said above, the problem with using the spec is that we use the date of the end day to set the days that the event appears. So in our current implementation, the example I gave would appear on both 11/21 and 11/22. I expect this should be "easily" fixable by moving the logic that fixes that issue from the data model itself into the display code. Instead of subtracting a day from the end time of the event, in an all-day case we could just stop displaying it a day earlier than it says. I'll have a look at this.

I have some work on this already completed, but submitting a bug to keep track.

mcclurgm:
Fixes #580 

Changed to use the exclusive dtend for all day events (midnight after the last day it occurs) wherever possible. This fixes some logic errors with time comparisons and brings us up to spec with RFC 5545. To keep the same user-facing behavior, I added a new function to get inclusive times for all-day events, which we can use for user-facing display.


danrabbit:
@mcclurgm Can you resolve conflicts here?

mcclurgm:
Done

jeremypw:
@mcclurgm Sorry, conflicts have arisen again. If these are fixed, I'll try and review.

mcclurgm:
Aright. I'll check it out and hopefully have it done today.

mcclurgm:
@jeremypw I fixed the merge conflicts.

core/Tests/util-tests.vala
```diff
@@ -148,35 +220,52 @@ void test_daterange_all_day () {
...
     range = new Calendar.Util.DateRange (start_time, end_time);
+    // assert (Calendar.Util.icalcomponent_is_in_range (event, range));
+
+    // A range that shouldn't include the event end, but just barely (within
+    // timezone offset)
+    start_time = new DateTime.local (2019, 11, 22, 0, 0, 1);
+    end_time = new DateTime.local (2019, 11, 22, 12, 0, 0);
+    range = new Calendar.Util.DateRange (start_time, end_time);
+    assert (!Calendar.Util.icalcomponent_is_in_range (event, range));
+    // A range the should include the event
```
jeremypw, line 242: Should be: `// A range that should include the event`

jeremypw (dismissed):
This seems to behave similarly to master with respect to allday events as far as the display is concerned.  However, it does not fix the problem that the Agenda incorrectly shows an allday event on Monday as happening "Tomorrow" when the current day is Saturday  (tested in a UK locale and GMT timezone).  Not sure if this is in scope but it would be nice if it could be fixed while dealing with allday events and "event_in_range".

jeremypw:
Screenshot from 2020-11-29 06 14 17

Calendar containing Allday events created both with this PR and master on Sunday and Monday.  The current day is Saturday.

mcclurgm:
Hmm. I'm really low on time right now, so if this isn't a regression and this PR works as intended I'd rather merge it as is and return to that issue later. Otherwise it'll be at least a couple weeks before I can dedicate the time to tracking down and solving a new bug.

jeremypw:
@mcclurgm Fair enough.

@tintou Do you want to comment on this?

tintou (approved):
Looks good here

tintou:
Indeed, let's move forward, nice work :slightly_smiling_face:

PR #582, opened 2020-08-17
Make EventDialog modal

### LINKED ISSUE #370: Event Dialog is not modal
You can still interact with the calendar main window while the creating an event, including altering the active calendars. This does not seem desirable.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68306240-event-dialog-is-not-modal?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #370 

…in the most basic way. Literally just adds a `modal = true` property to the `EventDialog` window.
Is it really this simple? Would be great if it is, otherwise let me know and I'll work on the better solution.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Install Dependencies
       run: |
         apt update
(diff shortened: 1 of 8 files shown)


mcclurgm:
I think that's already taken care of elsewhere in the code. I just pushed a commit that makes this more explicit though.

PR #588, opened 2020-09-07
Calendar.Store: Abstract super class

marbetschar:
Another PR on our journey towards an unified API for Events and Tasks. This one provides a basic class structure as discussed in https://github.com/elementary/calendar/pull/560#discussion_r467212116.

### DIFF
--- a/core/Services/Calendar/Store.vala
+++ b/core/Services/Calendar/Store.vala
@@ -12,7 +12,9 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 
-public class Calendar.Store : Object {
+public abstract class Calendar.Store : Object {
+
+    public string? source_extension_name { get; construct; }
 
     /* The data_range is the range of dates for which this model is storing
      * data. The month_range is a subset of this range corresponding to the
@@ -59,22 +61,12 @@ public class Calendar.Store : Object {
     public GLib.Queue<E.Source> calendar_trash;
     private E.CredentialsPrompter credentials_prompter;
 
-    private static Calendar.Store? store = null;
-    private static GLib.Settings? state_settings = null;
+    protected static Calendar.Store? store = null;
+    protected static GLib.Settings? state_settings = null;
 
-    public static Calendar.Store get_default () {
-        if (store == null)
-            store = new Calendar.Store ();
-        return store;
-    }
+    protected Store (string source_extension_name) {
+        Object (source_extension_name: source_extension_name);
 
-    static construct {
-        if (SettingsSchemaSource.get_default ().lookup ("io.elementary.calendar.savedstate", true) != null) {
-            state_settings = new GLib.Settings ("io.elementary.calendar.savedstate");
-        }
-    }
-
-    protected Store () {
         this.week_starts_on = get_week_start ();
         this.month_start = Calendar.Util.datetime_get_start_of_month (get_page ());
         compute_ranges ();
@@ -98,9 +90,8 @@ public class Calendar.Store : Object {
             registry.source_added.connect (add_source);
 
             // Add sources
-            registry.list_sources (E.SOURCE_EXTENSION_CALENDAR).foreach ((source) => {
-                E.SourceCalendar cal = (E.SourceCalendar)source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
-                if (cal.selected == true && source.enabled == true) {
+            registry.list_sources (source_extension_name).foreach ((source) => {
+                if (is_source_active (source)) {
                     add_source (source);
                 }
             });
@@ -110,6 +101,8 @@ public class Calendar.Store : Object {
     }
 
     //--- Public Methods ---//
+
+    public abstract bool is_source_active (E.Source source);
 
     public void add_event (E.Source source, ECal.Component event) {
         add_event_async.begin (source, event);
@@ -241,8 +234,8 @@ public class Calendar.Store : Object {
         lock (source_client) {
             foreach (var id in source_client.get_keys ()) {
                 var source = registry.ref_source (id);
-                E.SourceCalendar cal = (E.SourceCalendar)source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
-                if (cal.selected == true && source.enabled == true) {
+
+                if (is_source_active (source)) {
                     load_source (source);
                 }
             }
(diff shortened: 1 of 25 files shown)


marbetschar:
@mcclurgm do you have any idea, why the tests are failing in this PR?

core/Services/Calendar/Store.vala
```diff
@@ -59,22 +61,12 @@ public class Calendar.Store : Object {
...
+    protected static Calendar.Store? store = null;
+    protected static GLib.Settings? state_settings = null;
 
-    static construct {
-        if (SettingsSchemaSource.get_default ().lookup ("io.elementary.calendar.savedstate", true) != null) {
-            state_settings = new GLib.Settings ("io.elementary.calendar.savedstate");
-        }
-    }
+    protected Store (string source_extension_name) {
+        Object (source_extension_name: source_extension_name);
```
mcclurgm, line 69: This seems to be the source of your issue with the test. I don't think anything after the `Object (...)` block is actually being run ([docs](https://wiki.gnome.org/Projects/Vala/Tutorial#GObject-Style_Construction)). Here's what fixed it for me:

```suggestion
    protected Store (string source_extension_name) {
        Object (source_extension_name: source_extension_name);
    }
    
    construct {
```

marbetschar: @mcclurgm good catch!! Coming from Java, I wasn't aware of this behaviour 🎉️

danrabbit:
I'm really hesitant to introduce this kind of abstraction when there is only one thing that subclasses it.

Historically, these things are never actually kept in sync across repositories and this ends up being lasagna. Everywhere that we'd tried to do that, we've ended up rewriting things to remove all the abstraction because it just makes it way harder to understand.

marbetschar:
@danrabbit what's the suggested way forward then? I'm running out of ideas here.

If I understand you correctly, is having a unified API for Tasks and Events not desireable as long as its code needs to be maintained across repositories - which I agree to. But on the other hand, regarding @tintou creating another library for this is not worth the effort.

So we end up where we started: Seperate code for handling Tasks or Events, residing in elementary/tasks respectively elementary/calendar. However, this approach makes it harder to include events and tasks in elementary/wingpanel-indicator-datetime.

jeremypw:
There are already some abstract classes in Granite - could it belong there? Or an interface definition.

mcclurgm:
If I can chime in here--if the issue is the abstract class, the reasoning is described a bit in the comments for the PR Marco linked. There will be 2 subclasses soon, as soon as we get task logic merged into this repo. In the last implementation of the Store, there was a fair amount of code with something like "if it's an event store, do this; otherwise, do that". This led to some deeply nested if statements and such. So we were thinking that moving the common implementation to a superclass and then having the event- or task-specific logic (that would otherwise be in an if statement) in subclasses would make it cleaner. I can see an argument either way though.

I'm not sure moving the abstraction to Granite makes much sense, since it's fairly specific to these apps (calendar, tasks, wingpanel).

jeremypw:
Well the existing Granite abstract classes are specific to the Settings app ...  I haven't looked into this in detail so I don't know whether the abstract class/interface could be made generic enough to be potentially useful for other (maybe 3rd party) apps referencing tasks and events.

marbetschar:
IMHO a shared base for use in further apps would definitely be nice - although for now it's only going to be used in Tasks, Calendar and Wingpanel.

marbetschar:
So, which architecture/approach does fit best in the overall elementary philosophy? And how should we get there? For a bit more context: The differences in source code for data handling in Calendar vs. Tasks comes down to just a few things. For Tasks we need to:

1. use a different [settings schema source path](https://github.com/marbetschar/calendar/blob/a622229bdd649ede36ad5ccf02f52522ceae9892/core/Services/Calendar/Store.vala#L79) if needed
2. [use the `E.SOURCE_EXTENSION_TASK_LIST` type instead of the `E.SOURCE_EXTENSION_CALENDAR` type](https://github.com/marbetschar/calendar/blob/a622229bdd649ede36ad5ccf02f52522ceae9892/core/Services/Calendar/Store.vala#L194-L197)
3. [use the `registry.default_task_list` instead of `registry.default_calendar`](https://github.com/marbetschar/calendar/blob/a622229bdd649ede36ad5ccf02f52522ceae9892/core/Services/Calendar/Store.vala#L295)
4. [use the `E.SourceTaskList` type instead of the `E.SourceCalendar` type](https://github.com/marbetschar/calendar/blob/a622229bdd649ede36ad5ccf02f52522ceae9892/core/Services/Calendar/Store.vala#L196)
5. [add an API which supports adding multiple views ("queries") to the same source](https://github.com/marbetschar/calendar/blob/a622229bdd649ede36ad5ccf02f52522ceae9892/core/Services/Calendar/Store.vala#L339),
otherwise [rendering a Task List per Source along with the Scheduled List does not work](https://github.com/elementary/tasks/pull/120#issue-421481685).
6. optionally: [add support to optimistically show the new state in the UI](https://github.com/elementary/tasks/pull/104#issuecomment-611206622) and revert it if the backend fails for some reason [to improve UX](https://github.com/elementary/calendar/pull/560#issuecomment-670646973) 

As you probably already know, I solved all of the outlined challenges above in a [proof of concept](https://github.com/marbetschar/calendar/blob/new-calendar-store/core/Services/Calendar/Store.vala). IMHO the discussion here is primarily about how we can move into the proof of concept direction without introducing regressions - because the overall needed change is quite huge.

**Evolution Data Server Jargon Definitions**

- Source = Calendar or Task List
- Component = Event or Task
- View = A live-query emitting signals (added, modified, removed) for Components which do match provided criteria within one Source
- Registry = A unified backend API, capable to return all availbale Sources

marbetschar:
... [digging into this a little deeper](https://valadoc.org/libedataserver-1.2/E.SourceRegistry.html), I'm wondering if its better to take this even further and build a generalised UX-optimized library for dealing with Calendar, Tasks, Mail and Contacts (e.g. an "EDS.Store").

... or if we should screw the whole "Calendar.Store" generalization approach and keep things on a per-App base, as they are now.

jeremypw:
Sounds an interesting idea but a fair amount of work, and requiring a new project.  Need @danrabbit to comment.

marbetschar:
If a library is the way to go, I do have the time to create and maintain it (with some help to kick things off). On the other hand, if we want to keep things on a per App base and avoid just another library I guess it's probably better to also not start mixing Code from Tasks and Calendar.

What do you think @danrabbit and @tintou ?

marbetschar:
talked to @tintou and @danrabbit - there is no need for another library. The root cause for the delay in the UI is because the EDS library is not used correctly. So we need to find a way to force it to sync asynchronously rather than synchronously.

PR #595, opened 2020-09-16
Calendar.Store => Calendar.EventStore

marbetschar:
Renamed Calendar.Store to Calendar.EventStore to make the similarities and differents between Tasks and Calendar more obvious and avoid further confusion. As stated in https://github.com/elementary/calendar/pull/588#issuecomment-692160733 there is currently no need to merge both Calendar.EventStore and Calendar.TaskStore using an abstract superclass. So we leave it as is for now.

This also superseds https://github.com/elementary/calendar/pull/560


tintou (approved):

tintou (changes requested):
You also need to update po/POTFILES

marbetschar:
@tintou good catch, thank you!

tintou (approved):

PR #599, opened 2020-09-22
Fix checkbox to set calendar as default

### LINKED ISSUE #550: Checking or unchecking one calendar as default should not automatically assign the same status to all other calendars
Screen record

"Birthdays & Anniversaries" and "Personal" are the standard calendars of Evolution in this example.
I use the latest stable and unmodified Hera version including all updates.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92263822-checking-or-unchecking-one-calendar-as-default-should-not-automatically-assign-the-same-status-to-all-other-calendars?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #550 

This keeps the state of the check button to set a calendar as default in the right state. When you open the calendar edit dialog, it always reverts to unchecked, and checking it will tell Calendar to make that calendar the default. If it's not checked, then nothing happens and the default calendar stays as is.

I'm not convinced this is a great design, but it at least fixes the issues we had before. It means that (outside of what appears when you create a new event) there's no indication of which is default. But as far as I can tell, this PR just fixes the current design, so any redesign can be done later.

If this gets approved I'll update appdata.xml then to avoid conflicts.

### DIFF
--- a/src/SourceDialog/SourceDialog.vala
+++ b/src/SourceDialog/SourceDialog.vala
@@ -47,7 +47,6 @@ public class Maya.View.SourceDialog : Gtk.Grid {
     public signal void go_back ();
 
     construct {
-        debug ("SourceDialog constructor");
         widgets_checked = new Gee.HashMap<string, bool> (null, null);
 
         var cancel_button = new Gtk.Button.with_label (_("Cancel"));
--- a/src/SourceDialog/SourceDialog.vala
+++ b/src/SourceDialog/SourceDialog.vala
@@ -23,7 +23,6 @@ public class Maya.View.SourceDialog : Gtk.Grid {
     public EventType event_type { get; private set; default=EventType.EDIT;}
 
     private Gtk.Entry name_entry;
-    private bool set_as_default = false;
     private string hex_color = "#da3d41";
     private Backend current_backend;
     private Gee.Collection<PlacementWidget> backend_widgets;
@@ -32,6 +31,7 @@ public class Maya.View.SourceDialog : Gtk.Grid {
     private Gtk.Button create_button;
     private Gtk.ComboBox type_combobox;
     private Gtk.ListStore list_store;
+    private Gtk.CheckButton is_default_check;
     private E.Source source = null;
 
     private Gtk.RadioButton color_button_red;
@@ -47,6 +47,7 @@ public class Maya.View.SourceDialog : Gtk.Grid {
     public signal void go_back ();
 
     construct {
+        debug ("SourceDialog constructor");
         widgets_checked = new Gee.HashMap<string, bool> (null, null);
 
         var cancel_button = new Gtk.Button.with_label (_("Cancel"));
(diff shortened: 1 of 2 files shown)


src/SourceDialog/SourceDialog.vala
```diff
@@ -246,13 +242,15 @@ public class Maya.View.SourceDialog : Gtk.Grid {
...
             create_button.set_label (_("Create Calendar"));
+            is_default_check.active = false;
         } else {
             event_type = EventType.EDIT;
             create_button.set_label (_("Save"));
             name_entry.text = source.display_name;
             type_combobox.sensitive = false;
             type_combobox.set_active (0);
             list_store.foreach (tree_foreach);
+            is_default_check.active = false;
```
danrabbit, line 253: Is this meant to be false on both sides of this conditional?

mcclurgm: No, it’s supposed to be this way. That’s what the last blurb in the main comment was about. It’s always unchecked, so I guess it’s something more like a “do I set this as default when the dialog exits“ check than a “default calendar” check. 

I don’t know of a good way to get the status of whether a calendar is default from here. Also, if I did make it reflect the default, then what would it mean to uncheck the current default calendar?

danrabbit: Ah okay. Yeah I guess it should probably reflect the correct state on open and then if it's checked we could make it insensitive so you can't uncheck, only check

mcclurgm: Oh, that’s a good idea. I’ll see what I can do!

mcclurgm: Alright, here's my take at implementing that. It's much nicer this way!

mcclurgm: One design question I suppose. Should I change the label to something like "Calendar is already the default" to show why it's insensitive, or do you think this is enough?

src/SourceDialog/SourceDialog.vala
```diff
@@ -250,9 +251,19 @@ public class Maya.View.SourceDialog : Gtk.Grid {
             type_combobox.sensitive = false;
             type_combobox.set_active (0);
             list_store.foreach (tree_foreach);
-            is_default_check.active = false;
-            var cal = (E.SourceCalendar)source.get_extension (E.SOURCE_EXTENSION_CALENDAR);
 
+            try {
+                var registry = new E.SourceRegistry.sync (null);
+                var source_is_default = source.equal (registry.default_calendar);
+                debug (@"source_is_default: $source_is_default");
```
danrabbit, line 258: Typically we use printf style:
```suggestion
                debug ("source_is_default: %b", source_is_default);
```

danrabbit (changes requested):
Seems to work as expected, just one tiny code style thing :)

src/SourceDialog/SourceDialog.vala
mcclurgm: Oops, that wasn't even supposed to be there. Deleted it.

src/SourceDialog/SourceDialog.vala
danrabbit: We can probably revisit that later if after some use it turns out it's confusing

danrabbit (approved):

mcclurgm:
Since you approved, while it's running the build, I'll update appdata real quick for this fix.

mcclurgm:
Done. Thanks for your suggestions!

4jNsY6fCVqZv:
@mcclurgm Thanks for solving the issue!

PR #605, opened 2020-10-06
Use color button styles from Greenfield

4jNsY6fCVqZv:
For reference see https://github.com/elementary/tasks/pull/129


fabiozaramella (approved):
Much better. Also, this makes them looks the same way as in the desktop plug not only with light theme but also in dark mode, so it's one step further in that direction.

fabiozaramella:
Marking as blocked waiting for a new release for Hera.

danrabbit (approved):
Lookin' good. nice work!

PR #607, opened 2020-10-08
Replace Granite.Widgets.Avatar with Hdy.Avatar

fabiozaramella:
Depends on:

- [x] https://github.com/elementary/stylesheet/pull/842

GIF of the new list in action:

Hdy Avatar in action in Calendar


src/EventEdition/GuestGrid.vala
```diff
@@ -106,8 +108,17 @@ public class Maya.View.EventEdition.GuestGrid : Gtk.Grid {
                         individual = map_iterator.get_value ();
                         if (individual != null) {
                             try {
-                                individual.avatar.load (32, null);
-                                avatar = new Granite.Widgets.Avatar.from_file (individual.avatar.to_string (), 32);
+                                individual.avatar.load (ICON_SIZE, null);
+                                avatar = new Hdy.Avatar (ICON_SIZE, individual.avatar.to_string (), true);
+                                avatar.set_image_load_func (() => {
```
tintou, line 113: You're not taking the scale facto into account here, I think that the intended behavior should be something like:
```suggestion
avatar.set_image_load_func ((size) => {
    try {
        return new Gdk.Pixbuf.from_stream_at_scale (individual.avatar.load (size, null), size, size, true);
    } catch (Error e) {
        debug (e.message);
        return null;
    }
});
```

fabiozaramella: Whops, you're right. Thanks for noticing. 😅️ I'll push these changes now.

danrabbit:
Moving to blocked pending one last release of Calendar for 5.x

meson.build
```diff
@@ -3,7 +3,12 @@ project('io.elementary.calendar',
     version: '5.1.1'
 )
 
-add_project_arguments('-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()), language: 'c')
+add_project_arguments([
+    '-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()),
+    '-DHANDY_USE_UNSTABLE_API'
+    ],
+    language: 'c'
+)
```
danrabbit, line 11: API is stable in 0.90.0

```suggestion
add_project_arguments('-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()), language: 'c')
```

src/EventEdition/GuestGrid.vala
```diff
@@ -106,8 +108,16 @@ public class Maya.View.EventEdition.GuestGrid : Gtk.Grid {
                         individual = map_iterator.get_value ();
                         if (individual != null) {
                             try {
-                                individual.avatar.load (32, null);
-                                avatar = new Granite.Widgets.Avatar.from_file (individual.avatar.to_string (), 32);
+                                individual.avatar.load (ICON_SIZE, null);
+                                avatar = new Hdy.Avatar (ICON_SIZE, individual.avatar.to_string (), true);
```
danrabbit, line 112: This doesn't look right. This would make the named fallback based on the file name. I don't think these two lines are actually necessary at all. We don't need to create a new avatar object here. In the old code this seems to be a workaround for shortcomings of Granite.Avatar

danrabbit (changes requested):
Nice work! This works as expected, but I think there's some unnecessary code. Left a few comments

src/EventEdition/GuestGrid.vala
tintou, line 112: Yes you might want to replace these lines with something like:
```suggestion
                                avatar = new Hdy.Avatar (ICON_SIZE, individual.display_name, true);
```

meson.build
fabiozaramella: Yep, you're right. Thanks for pointing out.

src/EventEdition/GuestGrid.vala
fabiozaramella: Yes, it was based on the old workaround for Granite.Avatar but doesn't makes sense here whilst the suggestion from @tintou does, according to Folks.Individual doc, so I think I'll apply his suggestion.

fabiozaramella: I've checked Folks.Individual doc again and I thinks that's correct. Thanks!

src/EventEdition/GuestGrid.vala
```diff
@@ -106,8 +108,15 @@ public class Maya.View.EventEdition.GuestGrid : Gtk.Grid {
                         individual = map_iterator.get_value ();
                         if (individual != null) {
                             try {
```
tintou, line 110: You don't need this `try{} catch(){}` clause anymore :slightly_smiling_face:

fabiozaramella: True. 😅 Sorry, I'm currently on mobile so I've applied your suggestions through web and I didn't notice. I'll update the PR as soon as i come back home. 👍🏻

fabiozaramella: Done.

src/EventEdition/GuestGrid.vala
```diff
@@ -105,12 +107,16 @@ public class Maya.View.EventEdition.GuestGrid : Gtk.Grid {
                     if (address.value == mail_address) {
                         individual = map_iterator.get_value ();
                         if (individual != null) {
-                            try {
-                                individual.avatar.load (32, null);
-                                avatar = new Granite.Widgets.Avatar.from_file (individual.avatar.to_string (), 32);
-                            } catch (Error e) {
-                                critical (e.message);
-                            }
+                            avatar = new Hdy.Avatar (ICON_SIZE, individual.display_name, true);
```
danrabbit, line 110: Since we've already created an avatar, we probably only need to update the text property:
```suggestion
                            avatar.text = individual.display_name;
```

danrabbit (changes requested):
Just one little suggestion, otherwise I think it looks good

src/EventEdition/GuestGrid.vala
fabiozaramella: I've applied the suggested change.

danrabbit (approved):
Looks good. Nice work!

Issue #479, opened 2019-12-27
Can't find ICS import dialog


CodingSpiderFox:
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Describe the bug
Cannot find the import ICS functionality.

### To Reproduce
Open elementary calendar, download an ICS file.
I don't see where I can import it.

### Expected behaviour
I'd like to have a menu bar with all the common calendar functions.

### Screenshots or screen recordings
![image](https://user-images.githubusercontent.com/25173078/71507628-0ea5ee80-287d-11ea-859b-19f2617c2e8d.png)

## Platform Information
ElementaryOS 5.1

Please check what applies:
-->
- [ ] I'm using the latest version from git that I've manually compiled
- [X] I'm using the latest released stable version

mcclurgm:
I agree, this makes it hard to know that this functionality exists. This is more of a feature request than a bug though. 

To import an ICS file you have to open the file from Files, and then it'll open in Calendar and import. This makes sense in a way, but it's near impossible to know it's there. It would be very helpful if we could include some way of opening the pane from the Calendar UI. I think it would make sense in the calendar menu, next to the "Add Calendar..." button. Something like this, maybe:
![Screenshot from 2020-05-04 11 36 20](https://user-images.githubusercontent.com/29022469/80995530-9b4c5480-8dfb-11ea-8234-eb9663174467.png)

Issue #275, opened 2018-10-08
Can not import events to calendar


welaq:
On Juno Beta 2. I have an .ics file. In Files I right-click on .ics file -> Open with Calendar. And nothing happens. Just a couple days ago I did the same thing and everything worked fine.

sysfu:
Still having this issue on elementary OS 5.1.7 Hera. Please re-open. I can send an example file that fails to import.

PR #616, opened 2020-10-12
Import from ICS files

fabiozaramella:
This PR add the option to import calendars from ICS files from the gear menu, fixing #275 and fixing #479.

It also resolves some incorrect displaying of the number of calendars that are going to be imported.

### DIFF
--- a/src/ImportDialog.vala
+++ b/src/ImportDialog.vala
@@ -47,7 +47,7 @@ public class Maya.View.ImportDialog : Granite.MessageDialog {
                 "Events from %d file will be merged with this calendar:",
                 "Events from %d files will be merged with this calendar:",
                 files.length
-            );
+            ).printf (files.length);
         }
 
         calchooser_button = new Widgets.CalendarButton ();
--- a/src/ImportDialog.vala
+++ b/src/ImportDialog.vala
@@ -41,7 +41,7 @@ public class Maya.View.ImportDialog : Granite.MessageDialog {
                 critical (e.message);
             }
 
-            secondary_text = _("Events from \"%s\" will be merged with this calendar:").printf (name);
+            secondary_text = _(("Events from \"%s\" will be merged with this calendar:").printf (name));
         } else {
             secondary_text = ngettext (
                 "Events from %d file will be merged with this calendar:",
--- a/src/ImportDialog.vala
+++ b/src/ImportDialog.vala
@@ -41,12 +41,12 @@ public class Maya.View.ImportDialog : Granite.MessageDialog {
                 critical (e.message);
             }
 
-            secondary_text = _(("Events from \"%s\" will be merged with this calendar:").printf (name));
+            secondary_text = _("Events from \"%s\" will be merged with this calendar:").printf (name);
         } else {
             secondary_text = ngettext (
                 "Events from %d file will be merged with this calendar:",
                 "Events from %d files will be merged with this calendar:",
-                files.length
+                (ulong) files.length
             ).printf (files.length);
         }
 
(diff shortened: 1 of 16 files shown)


tintou:
FYI, [we should stop using Gtk.Dialog(Native).run](https://developer.gnome.org/gtk4/unstable/ch41s02.html#id-1.7.4.4.75) as it is removed from Gtk4, we should use dialog.response.connect () with dialog.show()

fabiozaramella:
> FYI, [we should stop using Gtk.Dialog(Native).run](https://developer.gnome.org/gtk4/unstable/ch41s02.html#id-1.7.4.4.75) as it is removed from Gtk4, we should use dialog.response.connect () with dialog.show()

Thank you for this! I'll update the PR then. 👍🏻

src/SourceDialog/SourceSelector.vala
```diff
@@ -71,6 +75,43 @@ public class Maya.View.SourceSelector : Gtk.Popover {
             create_source ();
             return Gdk.EVENT_STOP;
         });
+
+        import_calendar_button.button_release_event.connect (() => {
+            GLib.File[] files = null;
```
tintou, line 80: That can go in the `response` block

src/SourceDialog/SourceSelector.vala
```diff
@@ -71,6 +75,43 @@ public class Maya.View.SourceSelector : Gtk.Popover {
...
+            file_chooser.set_local_only (true);
+            file_chooser.set_select_multiple (true);
+            file_chooser.set_filter (ics_filter);
+
+            file_chooser.show ();
+            this.hide ();
+
+            file_chooser.response.connect ((response_id) => {
+                if (response_id == Gtk.ResponseType.ACCEPT) {
+                    foreach (GLib.File selected_file in file_chooser.get_files ()) {
```
tintou, line 102:
```suggestion
                    foreach (unowned GLib.File selected_file in file_chooser.get_files ()) {
```

src/ImportDialog.vala
```diff
@@ -41,13 +41,13 @@ public class Maya.View.ImportDialog : Granite.MessageDialog {
                 critical (e.message);
             }
 
-            secondary_text = _("Events from \"%s\" will be merged with this calendar:").printf (name);
+            secondary_text = _(("Events from \"%s\" will be merged with this calendar:").printf (name));
```
tintou, line 44: This is wrong, the text has to be translated first then replaced

fabiozaramella: You're right, sorry. I've found the problem, someone mistranslated it in Italian using %d instead of %s... 🤦🏻‍♂️️

danrabbit (changes requested):
I guess this exposes that we're using pretty large row spacing here. Probably what we should do is remove row spacing and then add 3px top and bottom margin to the separator. That would look more natural for a menu

src/SourceDialog/SourceSelector.vala
```diff
@@ -51,13 +51,17 @@ public class Maya.View.SourceSelector : Gtk.Popover {
         var add_calendar_button = new Gtk.ModelButton ();
         add_calendar_button.text = _("Add New Calendar…");
 
+        var import_calendar_button = new Gtk.ModelButton ();
+        import_calendar_button.text = _("Import From ICS File…");
```
danrabbit, line 55: I guess we should probably use the user-facing name of the file format and not the extension

```suggestion
        var import_calendar_button = new Gtk.ModelButton () {
            text = _("Import iCalendar File…")
        };
```

fabiozaramella: Makes sense 👍🏼️

tintou (approved):

fabiozaramella:
> I guess this exposes that we're using pretty large row spacing here. Probably what we should do is remove row spacing and then add 3px top and bottom margin to the separator. That would look more natural for a menu

Is it better now? It looks ok to me with these margins.

danrabbit (approved):
Looks great, nice work!

Issue #628, opened 2020-10-28
Slow month switching with many events


mcclurgm:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
Also reported in #623, but reporting here to track separately.

When switching months with many events, it stutters and takes a long time to populate. It's over a second for me, on an i7 XPS 13, which should be fairly fast. I would say I have an average of 5 events per weekday, so that would be probably around 25-30 events total. When I page to these months, the animation doesn't play at all.

### To Reproduce
1. Create a lot of events in a single month, at least 3 or 4 per day (the number will vary by machine of course, but it doesn't need to be that many to see)
2. Switch to a different month and back.
3. Returning to the original month will be slow like I described. The animation won't play for a while, and in my case events keep populating after the switch.

### Expected behavior

Switching should be smooth. Ideally events would appear instantaneously, but that's less important.

### Logs
I tried a few times to page back and forth and run it through a profiler. Here are some results (note from different runs):
- [`gprof` results](https://gist.github.com/mcclurgm/b9275afdc8405d7275bda03d0a401fce)
- A `perf` run that I generated a flame graph for using Hotspot:
![Bildschirmfoto von 2020-10-27 19-15-33](https://user-images.githubusercontent.com/29022469/97379239-f4b4ae00-1889-11eb-82e5-2247073fe1cf.png)

## Platform Information

![about](https://user-images.githubusercontent.com/29022469/96930945-c1cd7d00-1479-11eb-93dd-ece8b52f048a.png)

- [x] I'm using the latest version from git that I've manually compiled (latest that supports Hera)
- [ ] I'm using the latest released stable version

## Additional context

Of course, this isn't something that can be "fixed" clearly, but I'm opening this to make it sorta trackable and hope my findings are useful to others. I'm currently working on reducing the work done by `calendar_util_component_is_on_day`, but this is not nearly enough to get it to run smoothly. There's going to be a fair bit more to be done until this runs quickly, especially on slower hardware.

PR #629, opened 2020-10-30
Store system timezone in EventStore

mcclurgm:
Helps #628 

I added a `system_timezone` property to `EventStore`, which we use instead of `ECal.Util.get_system_timezone`. I also keep it updated when the system timezone is changed according to DBus, so it should always be as correct as calling the function every time.

Some motivation for this, since it adds complexity: According to my profiling results, we're sinking a lot of time into `ECal.Util.get_system_timezone` of all things. When switching months with many events, this function takes up about 18.5% of all the app's time (according to the `prof` sampler), and about 3/4 of every call to `Calendar.Util.ecalcomponent_is_on_day`. Here's my flame graph, which I also attached in #628 (this is in the middle right).
Bildschirmfoto von 2020-10-27 19-15-33

So storing the result of this function (I figured `EventStore` was a reasonable place to do so) should significantly reduce the work we do when constructing a new grid view. In an example run, timing 136 events loading through `Grid.on_events_added`, the new version took only 4% as long as the original version on the wall clock. According to `get_monotonic_time`, the old version took about 3593398 microseconds while the new one took 169885.

(Sorry for pushing to the main repo, I messed up my git config)


mcclurgm:
Hmm, not sure what's going wrong in the tests. Converting to draft as I figure it out.

It looks like the call to `ECal.Util.get_system_timezone` is failing with an error, but I'm not sure how to debug that. Is it possible that something is missing on the CI server? It works fine on my local machine.

fabiozaramella:
@tintou do you know how to fix this?

mcclurgm:
Ok, it looks like the issue is that I can't connect to the DBus proxy for `org.freedesktop.timedate1`, and the resulting warning triggers the meson fail condition. I'm looking into [GTestDBus](https://valadoc.org/gio-2.0/GLib.TestDBus.html), but it's totally new to me so this might take me a while.

core/Services/Calendar/Util/DateTime.vala
```diff
@@ -57,10 +57,12 @@ namespace Calendar.Util {
                 result.zone = ICal.Timezone.get_builtin_timezone (timezone);
 #endif
             } else {
+                var event_store = Calendar.EventStore.get_default ();
+                unowned ICal.Timezone system_timezone = event_store.system_timezone;
```
jeremypw, line 61: This is not used

core/Services/Calendar/EventStore.vala
```diff
@@ -37,6 +42,9 @@ public class Calendar.EventStore : Object {
     /* The start of week, ie. Monday=1 or Sunday=7 */
     public GLib.DateWeekday week_starts_on { get; set; default = GLib.DateWeekday.MONDAY; }
 
+    /* The system time zone as an ICal.Timezone */
+    public unowned ICal.Timezone system_timezone {get; private set;}
```
jeremypw, line 46: Not sure you need to specifiy `unowned` as that is the default for properties.

core/Services/Calendar/EventStore.vala
```diff
@@ -79,6 +87,29 @@ public class Calendar.EventStore : Object {
...
+            SettingsManager settings_manager = Bus.get_proxy_sync (BusType.SYSTEM,
+                "org.freedesktop.timedate1", "/org/freedesktop/timedate1");
+            ((DBusProxy)settings_manager).g_properties_changed.connect ((changed, invalid) => {
+                var timezone = changed.lookup_value ("Timezone", GLib.VariantType.STRING);
+                if (timezone != null) {
+#if E_CAL_2_0
+                    this.system_timezone = ECal.util_get_system_timezone ();
+#else
+                    this.system_timezone = ECal.Util.get_system_timezone ();
+#endif
```
jeremypw, line 106: Would it be safer to take a copy of the timezone here and when it changes?

core/Services/Calendar/Util/ECalComponent.vala
```diff
@@ -46,11 +46,9 @@ namespace Calendar.Util {
     }
 
     public bool ecalcomponent_is_on_day (ECal.Component component, GLib.DateTime day) {
-#if E_CAL_2_0
-        unowned ICal.Timezone system_timezone = ECal.util_get_system_timezone ();
-#else
-        unowned ICal.Timezone system_timezone = ECal.Util.get_system_timezone ();
-#endif
+        var event_store = Calendar.EventStore.get_default ();
+        unowned ICal.Timezone system_timezone = event_store.system_timezone;
```
jeremypw, line 50: Do not really need this variable - reference `event_store.system_timezone` directly.

core/Services/Calendar/EventStore.vala
```diff
@@ -79,6 +87,29 @@ public class Calendar.EventStore : Object {
...
+#if E_CAL_2_0
+        this.system_timezone = ECal.util_get_system_timezone ();
+#else
+        this.system_timezone = ECal.Util.get_system_timezone ();
+#endif
+        // Listen for system timezone changes
+        try {
+            SettingsManager settings_manager = Bus.get_proxy_sync (BusType.SYSTEM,
+                "org.freedesktop.timedate1", "/org/freedesktop/timedate1");
+            ((DBusProxy)settings_manager).g_properties_changed.connect ((changed, invalid) => {
```
jeremypw, line 99: This does not seem to work (?). The signal handler is not triggered by changing the system timezone using `timedatectl` at the command line or otherwise (even though the wingpanel indicator time does change).  Not sure why or how to fix it :disappointed:

core/Services/Calendar/EventStore.vala
mcclurgm: Definitely agreed. Fixed.

core/Services/Calendar/EventStore.vala
mcclurgm: Thanks for your suggestions. I'll have a look at this and see what I can figure out.

tintou:
@mcclurgm Here is how I would do it:
https://github.com/elementary/calendar/commit/6755dd7e2374681115586bcb424b1d2fc0cb4c88

mcclurgm:
Hmm, I haven't looked at it very closely but on the surface it looks good. I like the changes you made to integrate this into `TimeManager`. I'll have a closer look at it once my vacation starts 🙂 

Thanks for working on that!

mcclurgm:
This looks good and it makes things much clearer. But it doesn't actually solve the issue with the properties_changed callback. It gets called for timezone changes, but the if clause where we capture the `Timezone` isn't called. I'm having a hard time figuring out Variants, but here's the result of printing the variant when a timezone is changed: `{'Timezone': <'America/Phoenix'>}`.

Once we have that working, I have a couple changes to (hopefully) make this more readable, but first things first.

tintou:
@mcclurgm How do you figure-out that `g_properties_changed` isn't working on your end? On my side I always go inside the `if (timezone != null) {` when I change my timezone

mcclurgm:
I'm setting a debug message in the if clause, and it's not printing. Here's what the code is:
```vala
    private void on_timedate_properties_changed (Variant changed_properties, string[] invalidated_properties) {
    
        debug ("on_timedate_properties_changed");
        debug (changed_properties.print (true));
        var timezone = changed_properties.lookup_value ("Timezone", GLib.VariantType.VARIANT);
        
        if (timezone != null) {
#if E_CAL_2_0
            this.system_timezone = ECal.util_get_system_timezone ();
#else
            this.system_timezone = ECal.Util.get_system_timezone ();
            debug (@"Timezone changed: $timezone");
#endif
        }
        ...
```
I open calendar with debug messages enabled, and then go to settings and change the timezone. The first 2 debugs in that code print, but the third in the if clause doesn't.

mcclurgm:
Ok, I realized that I was trying to print in the wrong  `#if` block. But still, if I put a debug statement outside the ifs (right before the closing brace in the code I pasted), it still doesn't print. So I'm still not getting it, and I think I should be now.

mcclurgm:
Alright @tintou I merged in your changes. Thanks a lot! 

I made some small tweaks. Copied the timezone as @jeremypw suggested a while ago. I also deleted the `time_usec` variable from the `FDO.TimeDate1` interface. I didn't see that property used anywhere, so thought I'd simplify. If there's a reason I should keep that around let me know and I'll add it back in.

tintou:
@mcclurgm Great, if you're happy with it you can remove the draft status :slightly_smiling_face:
And it's okay to not specify properties in the interface skeleton for D-Bus interfaces, so that works too.

mcclurgm:
Alright great. I would like to add some tests for this (`ecalcomponent_is_on_day`, for example). That would require figuring out `GTestDBus` though, otherwise it won't work on CI.

I could not do that, and merge this as it is. I could add those tests in a later PR if I figure out how it works. Thoughts anyone? If we think this is good as is without tests, then I'll go ahead and mark it as ready.

mcclurgm:
Well on second thoughts, I'm going to call this good. I can figure out the tests later, but it would be good to get this fix in and I don't think that's a real blocker. Marking as ready.

tintou (approved):

PR #631, opened 2020-11-05
Install icons

### LINKED ISSUE #103: don't reference and expect non-standard icons, ship branded icons instead

Marukesu:
fixes #103


danrabbit (changes requested):
It looks like there are also two instances in code:

https://github.com/elementary/calendar/blob/master/src/MainWindow.vala#L41

https://github.com/elementary/calendar/blob/master/daemon/Daemon.vala#L171

danrabbit (approved):

PR #635, opened 2020-11-08
Make EventStore ranges inclusive

### LINKED ISSUE #634: Events missing on last day of month grid
## What Happened

Events don't display on the last day of the month grid. Here's an example (all of these events repeat daily forever, so should happen on 12/6):
missing-end-of-month

It seems that we end the range of data in `EventStore` on the last day of the month: for example, Nov 30. But this means that any events on Nov 30 are out of range, and so aren't included in the model.

## Expected Behavior

All events that occur within the visible calendar range will appear.

## Steps to Reproduce

1. Create a new event on the last visible day in the grid. It can't be all day for this to happen.
2. Notice that it doesn't appear in the current month grid. If that day is visible in the next month grid, it will appear there.

### Logs

n/a

## Platform Information

Happens on latest release (Hera) and compiled from git (Ubuntu 20.04).

mcclurgm:
Fixes #634 

This PR makes it so that events occurring on the last day of the month and grid will properly appear in their respective ranges. I had to make some fixes elsewhere that assumed the old behavior to get this to work. I also added tests for these behaviors.

### DIFF
--- a/core/Services/Calendar/EventStore.vala
+++ b/core/Services/Calendar/EventStore.vala
@@ -349,18 +349,27 @@ public class Calendar.EventStore : Object {
         return dt;
     }
 
+    /** Set the values of month_range and data_range.
+     *
+     * month_range contains the entire month starting with month_start.
+     *
+     * data_range fills in the rest of the visible calendar block, including the
+     *  week before the month starts and the week after it ends.
+     */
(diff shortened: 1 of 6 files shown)


tintou (approved):

tintou:
With tests that's even better, thanks a lot!

Issue #492, opened 2020-02-24
Add Offline Support


marbetschar:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature

It would be awesome to have offline support in the Calendar app. Right now, if the app is started while we do not have any internet connection it does not display any event.

**Describe the solution you'd like**

Show all known events in the Calendar while in offline mode and sync possible modifications as soon as we are back online.

mcclurgm:
At least while it still requires a connection, it would be nice if it could show some sort of error message when it can't connect. As it is, it just displays an empty calendar when not connected, which is not helpful and potentially confusing.

fabiozaramella:
It seems that starting Calendar offline doesn't show previously imported online events yes, but only at first. In fact, after a while they show up properly, even offline, it's just ver slow. Can you reproduce that?

PR #636, opened 2020-11-11
Better offline support

fabiozaramella:
Helps with #492 by adding a warning to users when network is unavailable and by imediately loading cached events without waiting for backend as commented in the bug report.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -80,7 +80,7 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
         error_bar.get_content_area ().add (error_label);
 
         string title = _("Network Not Available.");
-        string details = _("While offline, a local copy of previously added calendars will be displayed.");
+        string details = _("Connect to the Internet to see additional details and new events from online calendars.");
 
         info_label = new Gtk.Label ("<b>%s</b> %s".printf (title, details));
         info_label.use_markup = true;
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -22,7 +22,6 @@
 public class Maya.MainWindow : Hdy.ApplicationWindow {
     public View.CalendarView calview;
     private View.AgendaView sidebar;
-    private GLib.NetworkMonitor network_monitor;
 
     public const string ACTION_PREFIX = "win.";
     public const string ACTION_NEW_EVENT = "action_new_event";
@@ -37,10 +36,9 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
     private static Gee.MultiMap<string, string> action_accelerators = new Gee.HashMultiMap<string, string> ();
 
     private Calendar.Widgets.HeaderBar headerbar;
+    private Calendar.Widgets.ConnectivityInfoBar info_bar;
     private Gtk.Label error_label;
-    private Gtk.Label info_label;
     private Gtk.InfoBar error_bar;
-    private Gtk.InfoBar info_bar;
 
     public MainWindow (Gtk.Application application) {
         Object (
@@ -79,20 +77,7 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
         };
         error_bar.get_content_area ().add (error_label);
 
-        unowned string title = _("Network Not Available.");
-        unowned string details = _("Connect to the Internet to see additional details and new events from online calendars.");
-
-        info_label = new Gtk.Label ("<b>%s</b> %s".printf (title, details));
-        info_label.use_markup = true;
-        info_label.wrap = true;
-
-        info_bar = new Gtk.InfoBar () {
-            message_type = Gtk.MessageType.WARNING,
-            revealed = false,
-            show_close_button = false
-        };
-        info_bar.get_content_area ().add (info_label);
-        info_bar.add_button (_("Network Settings…"), Gtk.ResponseType.ACCEPT);
+        info_bar = new Calendar.Widgets.ConnectivityInfoBar ();
 
         sidebar = new View.AgendaView ();
         sidebar.no_show_all = true;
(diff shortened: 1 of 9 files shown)


src/MainWindow.vala
```diff
@@ -59,18 +67,34 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
...
+            message_type = Gtk.MessageType.ERROR,
+            revealed = false,
+            show_close_button = true
+        };
+        error_bar.get_content_area ().add (error_label);
 
-        var infobar_label = new Gtk.Label (null);
-        infobar_label.show ();
+        string title = _("Network Not Available.");
+        string details = _("While offline, a local copy of previously added calendars will be displayed.");
```
cassidyjames, line 83: I think it's obvious that a cached copy is displayed. This is more similar to how we handle it in AppCenter, where we explain what connecting will let you do:

```suggestion
        string details = _("Connect to the Internet to see additional details and new events from online calendars.");
```

fabiozaramella: Thank you! 😊

src/MainWindow.vala
```diff
@@ -59,18 +67,34 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
...
+        error_bar = new Gtk.InfoBar () {
+            message_type = Gtk.MessageType.ERROR,
+            revealed = false,
+            show_close_button = true
+        };
+        error_bar.get_content_area ().add (error_label);
 
-        var infobar_label = new Gtk.Label (null);
-        infobar_label.show ();
+        string title = _("Network Not Available.");
```
tintou, line 82:
```suggestion
        unowned string title = _("Network Not Available.");
```

src/MainWindow.vala
```diff
@@ -59,18 +67,34 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
...
+        string details = _("Connect to the Internet to see additional details and new events from online calendars.");
```
tintou, line 83:
```suggestion
        unowned string details = _("Connect to the Internet to see additional details and new events from online calendars.");
```

tintou:
In an ideal world, commit 228c6eb9ccc816c30d0b45cc19d222ee1e68a14c would be merged separately, also I'd personally prefer having a Calendar.ConnectivityInfobar class that would hold the connectivity logic so that the MainWindow remains simple :slightly_smiling_face:

fabiozaramella:
> In an ideal world, commit [228c6eb](https://github.com/elementary/calendar/commit/228c6eb9ccc816c30d0b45cc19d222ee1e68a14c) would be merged separately, also I'd personally prefer having a Calendar.ConnectivityInfobar class that would hold the connectivity logic so that the MainWindow remains simple 🙂

Makes sense 🙂 I'll update this PR and separate it from the MainWindow then. 👍🏻

src/MainWindow.vala
```diff
@@ -34,6 +35,11 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
     private uint configure_id;
     private static Gee.MultiMap<string, string> action_accelerators = new Gee.HashMultiMap<string, string> ();
 
+    private Calendar.Widgets.HeaderBar headerbar;
+    private Calendar.Widgets.ConnectivityInfoBar info_bar;
+    private Gtk.Label error_label;
+    private Gtk.InfoBar error_bar;
```
danrabbit, line 41: I'm not sure why the scope of all of these widgets have changed. It doesn't look like they are now accessed outside of the construct block. This should probably be reverted

src/Widgets/ConnectivityInfoBar.vala
```diff
@@ -0,0 +1,68 @@
...
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Fabio Zaramella
+ */
+
+public class Calendar.Widgets.ConnectivityInfoBar : Gtk.InfoBar {
+    private GLib.NetworkMonitor network_monitor;
+    private Gtk.Label info_label;
```
danrabbit, line 22: Doesn't look like these are used outside of `construct` so you could reduce scope here

src/Widgets/ConnectivityInfoBar.vala
```diff
@@ -0,0 +1,68 @@
...
+        );
+    }
+
+    construct {
+        unowned string title = _("Network Not Available.");
+        unowned string details = _("Connect to the Internet to see additional details and new events from online calendars.");
+
+        info_label = new Gtk.Label ("<b>%s</b> %s".printf (title, details));
+        info_label.use_markup = true;
+        info_label.wrap = true;
```
danrabbit, line 38: We could use latest code style here:
```suggestion
        info_label = new Gtk.Label ("<b>%s</b> %s".printf (title, details)) {
            use_markup = true,
            wrap = true
        };
```

src/Widgets/ConnectivityInfoBar.vala
```diff
@@ -0,0 +1,68 @@
...
+                    try {
+                        AppInfo.launch_default_for_uri ("settings://network", null);
+                    } catch (GLib.Error e) {
+                        critical (e.message);
+                    }
+                    break;
+                default:
+                    assert_not_reached ();
+            }
+        });
```
danrabbit, line 66: Since there is only one button, we can probably ignore response ID completely
```suggestion
        response.connect (() => {
            try {
                AppInfo.launch_default_for_uri ("settings://network", null);
            } catch (GLib.Error e) {
                critical (e.message);
            }
        });
```

danrabbit (changes requested):

fabiozaramella:
@danrabbit done 🙂️

src/Widgets/ConnectivityInfoBar.vala
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2011-2020 elementary, Inc. (https://elementary.io)
```
tintou, line 2:
```suggestion
 * Copyright 2020 elementary, Inc. (https://elementary.io)
```

tintou (approved):
It looks good from a code point of view, just change the copyright header and you're set

danrabbit (approved):

Issue #97, opened 2017-04-17
Add a option to start week on Monday


elementaryBot:
According to ISO 8601 the first day of a week is Monday. Many people prefer their calendars to start the week on Monday and most smartphone calendar applications (both in iOS and Android) offer this option.

Launchpad Details: [#LP1639843](https://bugs.launchpad.net/bugs/1639843) Antonio Camargo - 2016-11-07 15:54:02 +0000

fabiozaramella:
@elementary/ux similar to #269 where should be an appropriate place to toggle this setting?

PR #638, opened 2020-11-13
Use First Day of Week Setting from Switchboard Locale Plug

lainsce:
Using https://github.com/elementary/switchboard-plug-locale/pull/127, change the order of weekdays.

This one works live.


lainsce:
Oh, so I need to find a way to have the gschema from switchboard here... will I need to add Switchboard as a dependency?

lainsce:
I _think_ I found the error.

lainsce:
This PR does work, but CI build fails :(

mcclurgm:
Aw man, all those tests seem so great until they aren't. I've been getting issues with gsettings on CI in other branches too, so I'll need to figure something out for this. Sorry for messing up your PR haha...

lainsce:
> Aw man, all those tests seem so great until they aren't. I've been getting issues with gsettings on CI in other branches too, so I'll need to figure something out for this. Sorry for messing up your PR haha...

Nah, it's alright, there must be some CI bug out of reach.

mcclurgm:
I haven't looked at the error too closely (little time), but I had a thought today. If it's that the `io.elementary.switchboard.locale` schema doesn't exist, then I'd say we should account for that case in code, since it's not necessary for the app to function. You can [check that it does exist](https://developer.gnome.org/gio/stable/GSettings.html#g-settings-new), and we should be able to handle the case where it doesn't (default to the behavior we have right now, and don't set up the settings listener). 

Especially once we start distributing flatpaks, it would be good to handle cases where the user isn't on elementary. Based on the code we have, this looks like it should be relatively straightforward to implement. Famous last words, I know. But maybe just put some if statements around the code you added and that would do it.

lainsce:
I am already using `.lookup ()` method to see if it exists before loading it; the problem seems to be while building in CI testing.

mcclurgm:
Didn't notice that, sorry. I think I found the issue though. (Guess I don't want to get my other work done lol)

You check that the settings exists when you instantiate it, but never check again. Lines 95 and 345 both use `locale_settings` without ensuring it exists (isn't null).

lainsce:
Let's see it that fixes it.

lainsce:
Woohoo! Thanks @mcclurgm! That was the final piece.

core/Services/Calendar/EventStore.vala
```diff
@@ -330,6 +343,22 @@ public class Calendar.EventStore : Object {
...
+            if (week_start >= 1 && week_start <= 7) {
+                if (week_start_user_pref == week_start_posix) {
+                    week_start = week_start_posix + glib_offset;
+                } else {
+                    print ("User Preference for week start: %i\nSystem Preference for week start: %i\n", week_start_user_pref, week_start);
+                    week_start = week_start_user_pref + glib_offset;
+                }
+            }
+
+            return (GLib.DateWeekday) week_start;
```
mcclurgm, line 359: This isn't really necessary, since the next line is the same return anyway.

core/Services/Calendar/EventStore.vala
```diff
@@ -330,6 +343,22 @@ public class Calendar.EventStore : Object {
...
+        if (locale_settings != null) {
+            int week_start_user_pref = locale_settings.get_int ("first-day") + 1;
+
+            if (week_start >= 1 && week_start <= 7) {
+                if (week_start_user_pref == week_start_posix) {
+                    week_start = week_start_posix + glib_offset;
+                } else {
+                    print ("User Preference for week start: %i\nSystem Preference for week start: %i\n", week_start_user_pref, week_start);
+                    week_start = week_start_user_pref + glib_offset;
+                }
```
mcclurgm, line 356: I am a little concerned about this. Neither of these work use the Sunday special case. Also, doesn't line 351 just repeat the same calculation as line 341? Finally, this doesn't account for the fact that `glib_offset` isn't necessarily always the same. It depends on the value of the posix `week_1stday` variable. We need to take this into account either here or in switchboard. (All the locales I've found in Ubuntu use the same value, but that's not guaranteed.)

lainsce: You could tend to this in a newer PR, fixing both this here and in the datetime indicator.

core/Services/Calendar/EventStore.vala
lainsce: Both returns do things differently; one expects the locale setting to exist, and the other one just uses the non-locale setting calculations.

mcclurgm: Fair enough.

core/Services/Calendar/EventStore.vala
mcclurgm: I suppose so. I'm worried that it won't work as expected in this state though. Have you tried it set to Sunday? It looks to me like the Sunday bug should come into effect on Ubuntu systems. At least when the system locale is set to something like en_US.

If I'm wrong though, then I don't think this would ever impact an Ubuntu based system. I'd like to submit a PR later as you say, but it wouldn't be necessary to function.

lainsce: There's no apparent bugs if I set it to Sunday. :confused:

mcclurgm: Ok, seems like we're good then? I guess this works for me and I can come back around to the posix weirdness later when I have time.

lainsce: Alright. :)

mcclurgm:
This will need to be updated for the new changes in switchboard once they’re accepted.

mcclurgm:
This will also (mostly) fix #97. I think that's talking about a calendar-specific setting, but at least this gives the option to change it at all.

jeremypw:
Marking as "Blocked" pending merging of the the necessary switchboard setting.

PR #645, opened 2021-01-17
Embed calendar chooser in import dialog

### LINKED ISSUE #299: Import ICS dialog popover is cut off
screenshot from 2018-12-06 14 24 22 2x

We probably should just put the listbox directly in the dialog and not use a popover here

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67147892-import-ics-dialog-popover-is-cut-off?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #299 

This is a super work in progress, so I welcome feedback on the code or design! I'm submitting this PR now to get feedback. It's the first real Gtk work I've done, so I'd appreciate any suggestions. The design is also a first attempt.

Here's what it looks like:
import-dialog-popover

To do:
- [x] This needs some more testing, in case I messed something up when adapting `CalendarChooser`.
- [x] Should I use an embedded `CalendarChooser` in the event editor too?
- [x] If not, I should rewrite some of `CalendarButton` to reuse `CalendarChooser`.


src/Widgets/CalendarChooser.vala
```diff
@@ -0,0 +1,275 @@
...
+                search_entry.grab_focus ();
+            }
+        });
+
+        list_box.row_activated.connect ((row) => {
+            current_source = ((CalendarGrid)row.get_child ()).source;
+        });
+
+        list_box.selected_rows_changed.connect (() => {
+            foreach (var row in list_box.get_children ()) {
```
danrabbit, line 120: probably should make this an unowned ref
```suggestion
            foreach (unowned var row in list_box.get_children ()) {
```

danrabbit (changes requested):
Nice work breaking that out. Left a couple of small comments

> Should I use an embedded CalendarChooser in the event editor too? If not, I should rewrite some of CalendarButton to reuse CalendarChooser
Since space is at more of a premium in the event editor, it probably makes sense to stick with the popover/button design there. But yeah we should definitely DRY this so we don't have two classes that do a lot of the same work. It would make sense for `CalendarButton` to add a `CalendarChooser` to its popover

src/Widgets/CalendarChooser.vala
```diff
@@ -0,0 +1,275 @@
...
+        list_box.set_placeholder (placeholder);
+
+        var list_bg_css = "list { background: transparent; }";
+        try {
+            var css_provider = new Gtk.CssProvider ();
+            css_provider.load_from_data (list_bg_css);
+            list_box.get_style_context ().add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        } catch (Error e) {
+            warning ("Could not create CSS Provider: %s\nStylesheet:\n%s", e.message, list_bg_css);
+        }
```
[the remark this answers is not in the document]
mcclurgm: Wow, that's so much nicer. Thanks!

src/Widgets/CalendarChooser.vala
mcclurgm: I'll trust you on this. Need to learn more about Vala memory management... 🤓️

src/Widgets/CalendarChooser.vala
```diff
@@ -0,0 +1,275 @@
...
+                return child1.label.collate (child2.label);
+            } else {
+                return comparison;
+            }
+        });
+
+        list_box.move_cursor.connect ((mvmt, count) => {
+            var row = list_box.get_selected_row ().get_index ();
+            if (row == 0 && count == -1) {
+                search_entry.grab_focus ();
```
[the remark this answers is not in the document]
mcclurgm: Wouldn't this be more fragile, since it requires that the focus order never changes? If we always want to focus the search entry, it makes more sense to me to have it actively grab focus. But this isn't my code so I don't have a real connection to it 🙂️

mcclurgm:
Thanks for the feedback! I'll get working on the `CalendarButton`.

Before I get too far though, does this design look good to keep going? I'm not sure about the green background, or making the `ListBox` transparent. Also seems like far too much padding above the search field.

mcclurgm:
Alright, I've made what I think is an MVP for this, including `CalendarButton`. I think I've fixed all the regressions, at least the ones I've found so far.

One design question: in its current state, when the selected row is filtered out by the search and then comes back into view, it gets deselected. Should I add logic to re-select the active calendar row when it comes back into view (so it's always selected, except if navigating by keyboard)?

src/Widgets/CalendarChooser.vala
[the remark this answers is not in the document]
danrabbit: Well, theoretically, if say the search entry moved below the list for some reason, focusing the search entry when you move up would no longer be correct. But using `move_focus` whatever widget was there would be selected correctly. So it's more future proof, imo. But realistically this problem isn't an actual problem that will occur. Just minor nitpicking

src/Widgets/CalendarChooser.vala
```diff
@@ -0,0 +1,272 @@
...
+            header.margin_top = 0;
+        }
+    }
+
+    public void clear_search_entry () {
+        search_entry.text = "";
+    }
+}
+
+public class CalendarRow : Gtk.Grid {
```
danrabbit, line 199: If this needs to be a public class rather than a private class, it should have its own file

danrabbit (changes requested):
I made a small PR to your PR here: https://github.com/mcclurgm/calendar/pull/2

Also I left a comment about the CalendarRow class

Other than that, as far as I can tell this is working as expected :)

src/Widgets/CalendarChooser.vala
mcclurgm: Fair enough. Done. One question though: do you think that these CalendarChooser related widgets should go in a subdirectory, or is that just too many and it's fine as is?

danrabbit: It's probably okay as is. We can always reorganize if it gets out of hand :)

danrabbit (approved):
Alright, let's get this in! Thanks for fixing this issue :)

PR #656, opened 2021-03-07
Clickable links in event descriptions

### LINKED ISSUE #571: Links in comments aren't clickable
### Links in comments are plain text
screenshot
I often have to copy and paste long links from the comment field to the browser or other applications (Microsoft Teams).
Having clickable links would be a lot more comfortable.

igordsm:
Fix #571 . Implementation was based on [this mailing list message](https://mail.gnome.org/archives/gtk-app-devel-list/2008-March/msg00018.html). Requests for improvements are welcome. 

calendar-link-2

### DIFF
--- a/src/EventEdition/InfoPanel.vala
+++ b/src/EventEdition/InfoPanel.vala
@@ -264,6 +264,30 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
                 Gtk.TextBuffer buffer = new Gtk.TextBuffer (null);
                 buffer.text = property.get_comment ();
                 comment_textview.set_buffer (buffer);
+
+                var r = new GLib.Regex("https?://[\\w.-]*(\\/[\\w-\\d]*)*");
+                var link_tag = buffer.create_tag ("link");
+                link_tag.event.connect (on_link_clicked);
+                link_tag.set_property ("foreground", "#0000FF");
+                link_tag.underline = Pango.Underline.SINGLE;
+                GLib.MatchInfo mi;
+
+                Gtk.TextIter search_iter;
+                buffer.get_start_iter (out search_iter);
+                if (r.match(buffer.text, 0, out mi)) {
+                    do {
+                        int start, end;
+                        mi.fetch_pos (0, out start, out end);
+                        var link_text = buffer.text.substring(start, end - start);
+
+                        Gtk.TextIter starti, endi;
+                        search_iter.forward_search (link_text, Gtk.TextSearchFlags.TEXT_ONLY, out starti, out endi, null);
+
+                        buffer.apply_tag (link_tag, starti, endi);
+                        search_iter.forward_chars (link_text.length);
+                    } while (mi.next());
+                }
+
             }
 
             // Load the source
@@ -297,6 +321,28 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
             // Load the source
             calendar_button.current_source = parent_dialog.source;
         }
+    }
+
+    bool on_link_clicked (Gtk.TextTag tag, GLib.Object unused, Gdk.Event event, Gtk.TextIter iter) {
+        Gdk.ModifierType state;
+        event.get_state (out state);
+        var control = (state & Gdk.ModifierType.CONTROL_MASK) != 0;
+        if (event.type == Gdk.EventType.BUTTON_RELEASE && 
+            control) {
+            var button_event = (Gdk.EventButton) event;
+            if (button_event.button == Gdk.BUTTON_PRIMARY) {
+                var start = iter.copy ();
+                start.backward_to_tag_toggle (tag);
+
+                iter.forward_to_tag_toggle (tag);
+
+                var s = comment_textview.buffer.get_text (start, iter, false);
+                Gtk.show_uri_on_window (null, s, Gdk.CURRENT_TIME);
+                return true;
+            }
+        } 
+
+        return false;
     }
 
     Granite.Widgets.DatePicker make_date_picker () {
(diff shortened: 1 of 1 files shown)


src/EventEdition/InfoPanel.vala
```diff
@@ -264,6 +285,31 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
...
+                        mi.fetch_pos (0, out start, out end);
+                        var link_text = buffer.text.substring (start, end - start);
+
+                        Gtk.TextIter starti, endi;
+                        search_iter.forward_search (link_text, Gtk.TextSearchFlags.TEXT_ONLY, out starti, out endi, null);
+
+                        buffer.apply_tag (comment_link_tag, starti, endi);
+                        search_iter.forward_chars (link_text.length);
+                    } while (mi.next ());
+                }
```
pongloongyeat, line 311: I'm not familiar with this so please don't take my comment too seriously. Would it be possible to do the regex search on `TextBuffer.changed`? I tried your PR and when I paste links in, it only adds the tag after I've saved the event and reopened it. Note however that doing a regex search on each `TextBuffer.changed` would probably be incredibly inefficient but I am of the opinion that event descriptions are typically short and at most contain 100 words or so so performance shouldn't be too big of a hit.

mcclurgm: I don't think we should assume that descriptions are going to be short, especially the type that have links. Auto-generated Zoom or Webex events have very long descriptions. That said, I'm not sure what the performance hit of regex on change would be.

Two potential compromises I can think of:
1. Run the regex when the description field loses focus (so the focus goes somewhere else, like the title). This still wouldn't parse immediately, but it's possible that people could change the description and then go somewhere else, and then they'd see the benefit while they're still editing.
2. Use a timeout to wait a second or two after editing stops, then parse. I've seen some work over in elementary/code that worked with this.

mcclurgm:
Thanks for the PR! It seems to work as advertised for me. I haven't taken a very close look at the code, but in the meantime there are a few features that could use some work here:

1. The formatting of the link should follow the system theme. I'm not sure how to get this to work in a TextView, but it should use the current accent color and look good in both dark and light modes. @elementary/ux do we want to use the system accent color for links like this or do we want to use the green accent that we use elsewhere in Calendar?
2. There should be some sort of tooltip to indicate that links are clickable. Include something like "Hold ctrl and click to follow link" (the copy could use some work, of course).

src/EventEdition/InfoPanel.vala
```diff
@@ -264,6 +285,31 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
                 Gtk.TextBuffer buffer = new Gtk.TextBuffer (null);
                 buffer.text = property.get_comment ();
                 comment_textview.set_buffer (buffer);
+
+                // Identify links in the description and apply a tag to style them
+                var link_regex = new GLib.Regex ("https?://[\\w.-]*(\\/[\\w-\\d]*)*");
```
mcclurgm, line 290: This doesn't account for some types of URLs, for example ones that include `?` parameters. This is common in things like Zoom links. 

[Here's a blog post](https://mathiasbynens.be/demo/url-regex) that goes over some example patterns. I'm not sure if there's any licensing issues involved with copying these, however.

mcclurgm, line 290: The compiler also gives a warning of an unhandled `GLib.RegexError` on this line.

src/EventEdition/InfoPanel.vala
```diff
@@ -264,6 +285,31 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
...
+                        int start, end;
+                        mi.fetch_pos (0, out start, out end);
+                        var link_text = buffer.text.substring (start, end - start);
+
+                        Gtk.TextIter starti, endi;
+                        search_iter.forward_search (link_text, Gtk.TextSearchFlags.TEXT_ONLY, out starti, out endi, null);
+
+                        buffer.apply_tag (comment_link_tag, starti, endi);
+                        search_iter.forward_chars (link_text.length);
+                    } while (mi.next ());
```
mcclurgm, line 310: The compiler gives a warning of an unhandled `GLib.RegexError` here.

src/EventEdition/InfoPanel.vala
```diff
@@ -299,6 +345,27 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
...
+        if (event.type == Gdk.EventType.BUTTON_RELEASE &&
+            control) {
+            var button_event = (Gdk.EventButton) event;
+            if (button_event.button == Gdk.BUTTON_PRIMARY) {
+                var start = iter.copy ();
+                start.backward_to_tag_toggle (tag);
+                iter.forward_to_tag_toggle (tag);
+
+                var link_dst = comment_textview.buffer.get_text (start, iter, false);
+                Gtk.show_uri_on_window (null, link_dst, Gdk.CURRENT_TIME);
```
mcclurgm, line 361: This also has an unhandled error of some sort.

mcclurgm (changes requested):
The regex that you're using isn't quite complete and could use some updating. It misses some common cases. I also noticed when compiling on my machine that there are a few unhandled errors that should be taken care of.

danrabbit:
I'm gonna convert this back to draft since there hasn't been any movement in the last couple weeks. Feel free to mark ready for review once requested changes have been made :)

igordsm:
I'm waiting on the granite update to proceed. As soon as it lands I'll get back to this.

teamcons:
This PR is long stale and has been superseded by another merged PR. I would like to request closing it, to leave more breathing room to actual, relevant PR ?

PR #669, opened 2021-05-28
Define all-day events in the local timezone and document

mcclurgm:
This is largely an attempt to make this behavior well-defined, which I don't think it was previously.

Since `DATE`-type times are kind of implicitly floating in iCalendar, it makes sense to treat them in local time, since that's how floating times behave. This also removes a fair amount of special-casing for all-day events (especially those with `DATE`-type times), since they're now treated just like other events now.

It also fixes a bug with events that are missing both `DTEND` and `DURATION`, which is tied to all-day handling.

I also documented this definition in `datetime_is_all_day`, which will hopefully help clear up confusion. At least I know I was confused in the past about this definition.


jeremypw (approved):
Makes sense to me!

PR #670, opened 2021-06-05
Location entry, repeat switch grab focus by default

mcclurgm:
Just something I noticed when looking around the editor panel this evening. I think this is a better design, but feel free to disagree. It didn't take me very long to change! 🙂 

As is, the location and repeat panels have strange widgets focused by default. I don't know what the location panel focuses exactly, but it's not the entry, which I would imagine is the only widget users would want to interact with. Similarly, when an event is repeating, the "Add exception" button is focused by default in the repeat panel. I doubt this is used much, so I think that focusing the on/off switch makes more sense. It's both at the very start of the pane, making it a reasonable place to default to, and I assume the most predictably common action in the pane. (If you think there's a different place in that pane that should be selected by default, I'd be happy to change it to that!)

I'm not entirely sure how to set default focus, since I'm relatively new to working on GUIs in GTK. Let me know if there's a better way to do it, this is just my best guess.


danrabbit (approved):
Good call!

PR #673, opened 2021-06-09
Use Granite small label class for AgendaEventRow

mcclurgm:
New:
calendar-after
Old:
calendar-before

I also did some code style updates.


src/Widgets/AgendaEventRow.vala
```diff
@@ -342,7 +344,7 @@ public class Maya.View.AgendaEventRow : Gtk.ListBoxRow {
             }
         }
 
-        datatime_label.label = "<small>%s</small>".printf (datetime_string);
+        datetime_label.label = "%s".printf (datetime_string);
```
ryonakano, line 347:
```suggestion
        datetime_label.label = datetime_string;
```

Maybe we can set datetime_string directly into the label string, right?

danrabbit (approved):
Nice work

Issue #676, opened 2021-06-30
Event alerts are not working


andirsun:
## What Happened

I set 3 alerts for a testing event and none of them trigger an notification or sound.

**EDIT: Only the alert "At time of the event show"**

## Expected Behavior

Send a notification with the event info

## Steps to Reproduce

1. Create event
2. Set alerts 
3. Alerts dont work

### Logs

![Screenshot from 2021-06-30 08-27-39](https://user-images.githubusercontent.com/22928302/123968964-5ed3a480-d97d-11eb-88d2-960ec917e13b.png)

shawn1thompson:
I am having the same issue. When would I expect to see it fixed seeing as it's closed? Thank you.

PR #681, opened 2021-07-20
Flatpak: Create Autostart File Via Background Portal

Marukesu:
with this calendar will call the background portal to create the daemon autostart on the first launch when sandboxed.


danrabbit:
Is this a fix for #676 ?

Marukesu:
Sadly no, this only allow us to create the autostart file when installed as flatpak. since the one that comes with the build is useless on flatpak.

Since gala don't implement the `org.gnome.Shell.Introspection` interface, the background portal do nothing at all.

marbetschar:
@Marukesu FYI I added a PR which unifies the daemon and regular app into the same binary. So we may have to rework this a little once this is merged: https://github.com/elementary/calendar/pull/702

danrabbit:
@Marukesu can you update this/resolve conflicts since there's now a single binary?

Marukesu:
Updated the PR, now this will show the dialog in two cases:
* during launch with `--background`.
* when closing the window.

Also, the daemon is always started now, so that we can show notifications when the application is open, even if we don't have permission to run in background.

danrabbit:
Is there a reason to only use the background portal when running in Flatpak? Can we use it all the time?

Marukesu:
> Is there a reason to only use the background portal when running in Flatpak? Can we use it all the time?

The GTK portal backend uses the sandboxed app id to name the autostart file, but for host apps  it is empty, so they create an unnamed .desktop that is replaced by others application on host that calls the portal (see flatpak/xdg-desktop-portal#650).

data/io.elementary.calendar.gschema.xml
```diff
@@ -42,5 +42,10 @@
             <summary>The x coordinate of the hpaned dragger.</summary>
             <description>The x coordinate of the hpaned dragger.</description>
         </key>
+        <key name="background" type="b">
+            <default>false</default>
+            <summary>Should keep on background (Flatpak only)</summary>
```
jeremypw, line 47:
```suggestion
            <summary>Should keep running in the background (Flatpak only)</summary>
```

src/Application.vala
```diff
@@ -169,6 +176,40 @@ namespace Maya {
...
+            if (can_background || !FileUtils.test ("/.flatpak-info", FileTest.EXISTS)) {
+                return true;
+            }
+
+            string[] cmd = { "io.elementary.calendar", "--background" };
+            string handle = active_window != null ? yield ((MainWindow) active_window).export () : "";
+
+            var options = new HashTable<string, Variant> (str_hash, str_equal);
+            options["handle_token"] = Portal.generate_token ();
+            options["reason"] = _("Calendar wants to run on background and initialize with the session for events notifications");
```
jeremypw, line 190:
```suggestion
            options["reason"] = _("Calendar wants to run in the background and initialize with the session for events notifications");
```

src/MainWindow.vala
```diff
@@ -182,4 +183,51 @@ public class Maya.MainWindow : Hdy.ApplicationWindow {
...
+        ((Application) application).ask_for_background.begin ((obj, res) => {
+            unowned var app = (Application) obj;
+            loop.quit ();
+
+            if (!app.ask_for_background.end (res)) {
+                app.release ();
+            }
+        });
+
+        if (loop.is_running ()) {
```
jeremypw, line 198: Is this right?  It looks like you are running a loop that is already running?

jeremypw (changes requested):
I can confirm that after removing the existing Calendar installation with AppCenter and renaming the existing autostart file and installing a flatpak built from this PR and logging out and in again the following:
1. No daemon was running initially
2. Running the flatpak with `flatpak run` launched the program but did not start the daemon
3. Closing the window triggered a dialog requesting permission to run in background
4. After giving permission the window closed but the daemon did not start
5. After logging out and back in, the daemon did autostart.
6. It was not possible to launch Calendar from the wingpanel indicator but launching from the dock and Applications Menu worked.  (Is this an existing bug with flatpak/wingpanel?)

I would expect the daemon to start after giving it permission to do so and not have to log out and in.

src/MainWindow.vala
Marukesu: Yes, we use the loop to block the function while `application.ask_for_background` is running, so we should only call `loop.run()` if `is_running` is true.

The need to block the function here is that we don't delete the application window while the portal dialog is open.

Marukesu:
> 2. Running the flatpak with `flatpak run` launched the program but did not start the daemon

I can't reproduce this here, starting the application should start the daemon, the only diference from master here is that starting the daemon is done before the `--background` check.

> 4. After giving permission the window closed but the daemon did not start

The last commit should fix that, for some reason, chaining up `base.delete_event ()` was causing an segfault.

> 6. It was not possible to launch Calendar from the wingpanel indicator but launching from the dock and Applications Menu worked.  (Is this an existing bug with flatpak/wingpanel?)

Looking at the code in the indicator, it tries to launch `io.elementary.calendar --show-day`, so i don't expect it to work with the sandboxed calendar.

danirabbit:
@Marukesu wanna update this again and remove the flatpak check since the background portal can work with host apps now? :)

src/MainWindow.vala
Marukesu: Updated this part to stop the event handling instead of waiting the method response to continue.

Marukesu:
@danirabbit this should be good to review again.

src/Application.vala
```diff
@@ -166,6 +173,36 @@ namespace Maya {
...
         }
 
+        public async bool ask_for_background () {
+            string handle = active_window != null ? yield ((MainWindow) active_window).export () : "";
+            string[] cmd = { "io.elementary.calendar", "--background" };
+            bool can_background = true;
+
+            var options = new HashTable<string, Variant> (str_hash, str_equal);
+            options["handle_token"] = Portal.generate_token ();
+            options["reason"] = _("Calendar wants to run in the background and initialize with the session for events notifications");
```
danirabbit, line 183: More in line with the copy we just merged to Mail

```suggestion
            options["reason"] = _("Calendar will automatically start when this device turns on and run when its window is closed so that it can send event notifications.");
```

danirabbit (changes requested):
Seems like it would simplify things quite a bit to use LibPortal like we did for Mail?

src/Application.vala
```diff
@@ -61,10 +63,17 @@ namespace Maya {
         }
 
         protected override void activate () {
+            new Calendar.TodayEventMonitor ();
```
danirabbit, line 66: Should this be done in `activate`? it looks like Mail does the equivalent in `startup`

Marukesu: Well, Calendar doesn't have a `startup` method. This work since TodayEventMonitor is now marked as a `[SingleInstance]` class and the construct method will only be called one time.

Marukesu: Well, Calendar doesn't have a `startup` method. This work since TodayEventMonitor is marked as a `[SingleInstance]` class and the construct method will only be called one time.

danirabbit (approved):
Let's go :rocket: Thanks for working on this!

PR #683, opened 2021-07-20
Daemon: Use `Gee.HashMultiMap`

### LINKED ISSUE #676: Event alerts are not working

Marukesu:
this makes the daemon stores all the alarms for a event instead of only one.

fixes #676

### DIFF
--- a/daemon/Daemon.vala
+++ b/daemon/Daemon.vala
@@ -26,12 +26,12 @@ namespace Maya {
     };
 
     public class Daemon : GLib.Application {
-        private Gee.HashMap<ECal.Component, string> event_uid;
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Makes sense to me! Thanks for looking into this :)

Issue #674, opened 2021-06-10
Events have wrong timezone


bobarthur:
Im using Evolution Mail with calendar events and appointments from a WebDAV server. I create and edit the calendar events on several devices (Android Smartphone, Android Tablet, Outlook on Windows, Evolution Mail on Linux). The events are shown in the correct time across all devices, except in the Elementary Calendar when I click on the time shown in the top panel.

Example: Appointments which are created in Evolution Mail in my timezone GMT+1 at 8am show as 10am in the Elementary Calendar.

mcclurgm:
Thanks for submitting this! I have some questions to narrow this down.

First, which version are you using? Are you on Hera (the current release), or the Odin beta?

You mentioned this happens in the top panel, does it also happen in the main calendar app or is it just the top panel?

If you don't mind, could you attach an export of one of the events that displays wrong? You can export an ICS file from Evolution, and that would be a big help to tracking this down.

mcclurgm:
Could you upload an ICS file of one of the events that's displayed wrong? You can either link a file or just paste its contents in a comment. There are many possible issues that can cause this, so it would be really helpful for figuring out what's the problem for you.

PR #687, opened 2021-08-11
l10n: Correctly setup the locales

bobby285271:
Provides the directory where the locales are actually installed.

We are [packaging this](https://github.com/NixOS/nixpkgs/pull/130380#issuecomment-895720580) in NixOS, and due to NixOS 's special `localedir`, we cannot apply the translations without this patch.

Thanks in advance for reviewing this :-)


dmanlfc:
Will this PR fix - https://github.com/elementary/calendar/issues/674 ???

ryonakano (approved):

tintou (approved):

PR #689, opened 2021-08-12
Strip Alpha From Calendar Color if Needed

Marukesu:
Gtk.CssProvider fails if when setting `accent_color` to a RGBA value, this PR make sure that we are setting  only the RGB part of the hex color if we recevie a RGBA one.

### DIFF
--- a/src/Grid/EventButton.vala
+++ b/src/Grid/EventButton.vala
@@ -123,7 +123,7 @@ public class Maya.View.EventButton : Gtk.Revealer {
     private void reload_css (string background_color) {
         var provider = new Gtk.CssProvider ();
         try {
-            var colored_css = EVENT_CSS.printf (background_color.length == 9 ? background_color[0:-2] : background_color);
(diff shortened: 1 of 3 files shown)


src/Grid/EventButton.vala
```diff
@@ -123,7 +123,7 @@ public class Maya.View.EventButton : Gtk.Revealer {
     private void reload_css (string background_color) {
         var provider = new Gtk.CssProvider ();
         try {
-            var colored_css = EVENT_CSS.printf (background_color);
+            var colored_css = EVENT_CSS.printf (background_color.length == 9 ? background_color[0:-2] : background_color);
```
danrabbit, line 126: It looks like there's precedent for this in Tasks and it's done with `slice`:
```suggestion
            var colored_css = EVENT_CSS.printf (background_color.slice (0, 7));
```

danrabbit (changes requested):
Thanks for fixing this! It looks like this can also be achieved with `slice` and it's just a bit less verbose

danrabbit (approved):

PR #699, opened 2021-08-28
Add delete confirmation dialog

### LINKED ISSUE #551: Confirm delete event action or provide undo or add undo
## Prerequisites

## Feature
**Is your feature request related to a problem? Please describe.**
Since Calendar doesn't provide an undo feature, destructive actions like deleting events are easy to do, and don't provide any way to undo or confirm that this is what I wanted. An example of why this matters: today, I wanted to delete a single occurrence of a frequent event. I right-clicked on the event cell and quickly clicked "Remove event" instead of "Remove occurrence". This is easy to do since they're right next to each other. The result was I lost all of my past and future occurrences of that event, with no way to recover them. In general, we should provide either a confirmation or way to walk back destructive actions like this.

**Describe the solution you'd like**

Ideally, to follow the Elementary HIG, we should provide an option to undo these actions.
If that's not possible or we want something else in the meantime, then we should add a confirmation dialog, like you see in Files, for example.

**Existing work**

Here's an example of what a confirmation dialog could look like, from the iOS calendar app Fantasical:
fantastical

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92263821-confirm-delete-event-action-or-provide-undo-or-add-undo?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #551 

This is still in very WIP state, but I wanted to start this to get some design input.

This is what it looks like currently (sorry about the mix of German and English):
delete-dialog

What I would like design help on:

- Copy (for example, should I include the event title?)
- Should we wrap the deletion of a single occurrence into this dialog or keep it separate? Other calendar apps that I know combine the actions like this, but that doesn't mean we have to as well.

To do:

- [x] Account for recurring events in dialog text
- [x] Add to event context menu too
    - [x] Figure out how to make it transient to the main window
    - [x] Figure out how to use a single delete action throughout editor dialog, context menus
- [ ] Wrap occurrence/event deletion into dialog?
- [x] Add another dialog for deleting single occurrence if necessary


mcclurgm:
I believe this is ready for review. I would appreciate feedback on a few things:

- Copy: I think what I have is decent, but not great. It also conflicts with our use of "Remove" in the context menu.
- Does the widget class I made make sense, or should I just construct a custom dialog for each case? DRY, but it's used in few enough places that I could see it going either way.
- Is there a nicer way to construct the dialog, or should I keep everything in the "DeleteEventDialog" constructor block?

And of course, making sure that it works as expected and covers all cases. I tried deleting single and recurring events from both the event context menu and the event dialog, and also cancelling the dialog in all cases. If I've missed a way that we can delete events, I'll add them.

src/EventEdition/EventDialog.vala
```diff
@@ -214,9 +214,14 @@ public class EventDialog : Granite.Dialog {
...
         private void remove_event () {
-            var calmodel = Calendar.EventStore.get_default ();
-            calmodel.remove_event (original_source, ecal, mod_type);
-            this.destroy ();
+            assert (original_source != null);
+            var delete_dialog = new Calendar.DeleteEventDialog (original_source, ecal, mod_type) {
+                transient_for = this
+            };
+            var response = delete_dialog.run_dialog ();
+            if (response == Gtk.ResponseType.YES) {
```
jeremypw, line 222: It would be usual for the `delete_dialog` to be destroyed explicitly after getting the response, rather than trusting it will be destroyed when it goes out of scope.

src/EventEdition/EventDialog.vala
```diff
@@ -214,9 +214,14 @@ public class EventDialog : Granite.Dialog {
...
 
         private void remove_event () {
-            var calmodel = Calendar.EventStore.get_default ();
-            calmodel.remove_event (original_source, ecal, mod_type);
-            this.destroy ();
+            assert (original_source != null);
+            var delete_dialog = new Calendar.DeleteEventDialog (original_source, ecal, mod_type) {
+                transient_for = this
+            };
+            var response = delete_dialog.run_dialog ();
```
jeremypw, line 221: Its a bit unusual for the dialog to destroy itself? Normally the process that creates it does that.

src/Widgets/DeleteDialog.vala
```diff
@@ -0,0 +1,68 @@
...
+                delete_text = _("Delete Occurrence");
+            } else {
+                if (mod_type_prop != ECal.ObjModType.ALL) {
+                    warning (@"Creating delete dialog for unknown ObjModType: $mod_type_prop; defaulting to modify all");
+                }
+
+                if (summary == null) {
+                    title = _("Delete event?");
+                } else {
+                    title = _("Delete event “%s”?").printf (((!)summary).get_value ()); // Should question mark be in or out of quotes?
```
jeremypw, line 27: Question mark should not be in quotes so this is OK.

src/Widgets/DeleteDialog.vala
```diff
@@ -0,0 +1,68 @@
...
+    construct {
+        unowned var trash_button = add_button (delete_button_text, Gtk.ResponseType.YES);
+        trash_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+    }
+
+    public Gtk.ResponseType run_dialog () {
+        var response = (Gtk.ResponseType) this.run ();
+        if (response == Gtk.ResponseType.YES) {
+            var calmodel = Calendar.EventStore.get_default ();
+            calmodel.remove_event (source, ecal_event, mod_type_prop);
```
jeremypw, line 63: It seems strange for the dialog to take this action.  Normally the process creating the dialog takes the necessary action (?)

jeremypw:
Seems to work as expected.  I was a little concerned that the way dialogs were used might be unclear to someone coming new to the code but maybe that's just me.  I am not too concerned so if you think its OK I will approve.

mcclurgm:
Thanks for the review. I made a new class because I figured that this exact behavior would be used multiple places around the app, so I thought it made sense to package them together instead of replicating it. That's largely because it should run the same deletion logic and have the same text. If the practice I used here is strange, then I can break it up.

Maybe it would make sense to make a "delete" _action_ instead of a dialog? I'm not sure where this method would go, since there doesn't seem to be a class where it makes sense to me, but that would fix some of the strangeness here with constructing a subclass and the dialog destroying itself.

One more question I just realized: is there anywhere in the app that you can delete using a keyboard shortcut? I only changed the event dialog delete button and the two delete options in the context menu. I don't think there's one I'm missing, but I don't think so from a cursory search. The only other place I noticed `EventStore.remove_event` is in `MainWindow`, which I don't think is for a user action.

cassidyjames:
@mcclurgm 

>Should we wrap the deletion of a single occurrence into this dialog or keep it separate? Other calendar apps that I know combine the actions like this, but that doesn't mean we have to as well.

I'd like this (I came here to file it!) but it can be a follow-up since this alone would be a nice improvement. :smile:

jeremypw (approved):
Lets get this merged as it has UX approval and we can iterate on it if required.

Issue #677, opened 2021-06-30
Calendar notifications have not icon


andirsun:
## What Happened

When the notification of event is trigged then the icon shows but when is in the notification panel the icon is missing. 

![Screenshot from 2021-06-30 08-34-43](https://user-images.githubusercontent.com/22928302/123970047-68a9d780-d97e-11eb-990f-40fafe5028dd.png)

## Expected Behavior

Show an icon

## Steps to Reproduce

1.Create event
2. set an alert
3. wait the notification

Marukesu:
> I think the long fix is that we should be using the backgrounding portal stuff instead here

the background portal is not a solution for the non-sandboxed version, because it's generate a .desktop file using the received app-id and that is empty for non-sandboxed apps. we still need to create the autostart file manually in that case.

also, instead of creating a second .desktop file, is it possible to implement the daemon on the main binary like appcenter do?

marbetschar:
FYI: Addressed all the concerns in #702

marbetschar:
Closing since the corresponding PR has been merged.

PR #702, opened 2021-09-08
Embedded daemon in one unified binary

marbetschar:
This PR fixes the notification icon issue from #677 by embedding the daemonized background process into the same application binary (in the screenshot: "Other" is before this change, the correct notification is after this change):

Screenshot from 2021-09-08 08-52-28

This way, we gain the following benefits:

### 1. The notification icon is correct

The notification icon is read from the `Icon` key stored in `/usr/share/applications/{application_id}.desktop`. Since the background process shares the same `application_id` as the main application, the notification server uses the correct icon to display the alert.

### 2. The background process starts automatically

Autostart files are read from `/etc/xdg/autostart/{application_id}.desktop`. We can easily place a second desktop file with the EXEC line `io.elementary.calendar --background` in this directory.

### 3. When clicking on the notification, the corresponding app opens

When clicking on the notification, the notification server executes the `Exec` key of the `/usr/share/applications/{application_id}.desktop` file. Here it is crucial we embedded the background process within the same binary - because this way, the regular binary starts and the user is presented with the app itself. Just like we want it.

### 4. `System Settings > Notifications` work as expected

[As it turned out when working on this in Mail](https://github.com/elementary/mail/pull/679#issuecomment-914469849), if we don't embed the background process within the same binary we either end up with two application entries there - or even worse, with one which does not have any effect at all.


src/TodayEventMonitor.vala
```diff
@@ -0,0 +1,181 @@
...
+            ECal.ComponentAlarm e_alarm = event.get_alarm (alarm_uid);
+            ECal.ComponentAlarmAction action;
+
+#if E_CAL_2_0
+            action = e_alarm.get_action ();
+#else
+            e_alarm.get_action (out action);
+#endif
+            switch (action) {
+                case (ECal.ComponentAlarmAction.DISPLAY):
```
mcclurgm, line 85: This probably makes more sense as an if statement? There's currently only one case, so it would make sense that way as an if/else. But we can leave it as is to make future action types easier to add, if you want.

src/TodayEventMonitor.vala
```diff
@@ -0,0 +1,181 @@
...
+                        if (start_time.get_year () == now.get_year () && start_time.get_day_of_year () == now.get_day_of_year ()) {
+                            var time = time_until_now (start_time);
+                            if (time >= 0) {
+                                add_timeout.begin (source, event, (uint)time);
+                            }
+                        }
+                    }
+                    continue;
+                default:
+                    continue;
```
mcclurgm, line 113: Since the alarm has an action that we're not parsing, this would probably be a good case for a warning.

src/TodayEventMonitor.vala
```diff
@@ -0,0 +1,181 @@
...
+            if (extension != null) {
+                extension.set_last_notified (new DateTime.now_local ().to_string ());
+            }
+
+            queue_event_notification (event, uid);
+            return false;
+        });
+    }
+
+    public void queue_event_notification (ECal.Component event, string uid) {
```
mcclurgm, line 133:
```suggestion
    public void send_event_notification (ECal.Component event, string uid) {
```

I think this is a little clearer. The `add_timeout` function is really "queuing" the notification in the sense of adding it to a queue, while this function is called to actually send the notification.

src/TodayEventMonitor.vala
```diff
@@ -0,0 +1,181 @@
...
+                add_event (source, event);
+
+            return false;
+        });
+    }
+
+    private void on_events_updated (E.Source source, Gee.Collection<ECal.Component> events) {
+        Idle.add ( () => {
+            foreach (var event in events)
+                update_event (source, event);
```
mcclurgm, line 57: Rather strangely, I'm seeing that when I edit an event the `events` collection has size 0, so the callback is never added. I'm not sure quite how to debug this, so if you have any suggestions, let me know. In the meantime I'll keep digging and update if I find anything.

mcclurgm (changes requested):
I have some very small changes I noted. There's a bigger issue here that I noted for `on_events_updated`, but I'll come back with a second review when I can figure out what's going on there.

src/TodayEventMonitor.vala
mcclurgm: Never mind, I was seeing a combination of an EventStore bug and a user error about notifications. Ignore this.

mcclurgm:
Seems to work as expected. As I said in my last comment there are a few minor style things, but I think this pretty much works! I'll approve it once we get those sorted.

src/TodayEventMonitor.vala
marbetschar: Yeah, an if makes more sense here.

src/TodayEventMonitor.vala
marbetschar: Added a warning.

src/TodayEventMonitor.vala
marbetschar: Renamed the function.

marbetschar:
@mcclurgm thanks for your review! I addressed the proposed changes and I think we should be good to go here now.

mcclurgm:
Alright, seems good to me.

ghsttwn:
Latest Calendar version comes with the autostart entry for the daemon (at `/etc/xdg/autostart/io.elementary.calendar-daemon.desktop`) though it never really *starts* the daemon. 😕 Was this intended?

marbetschar:
@ghsttwn no, this was not the intention :)

The daemon should be autostarted as soon as you end your session and start a new one (read "logout and login again"). Doesn't this work on your end? If not, can you please open a new issue with a link to this PR, so we can investigate there?

PR #709, opened 2021-09-25
Do all timezone calculations in libical

### LINKED ISSUE #572: Windows timezones are not handled
I have connected my Office 365 calendar to Elementary Calendar application through public ics calendar.

**## Describe the bug**

ICS calendar content:
```
SUMMARY:Convite: IM0474 - Ecuador/Seidor Interface - PI - Semanal: de 10:00
  até 11:00 cada terça-feira de ter. 30 jun. até sáb. 15 ago. (BRT) ()
DTSTART;TZID=SA Pacific Standard Time:20200630T080000
DTEND;TZID=SA Pacific Standard Time:20200630T090000
CLASS:PUBLIC
PRIORITY:5
DTSTAMP:20200713T222111Z


SUMMARY:COMITE PARITARIO JULIO
DTSTART;TZID=SA Pacific Standard Time:20200714T150000
DTEND;TZID=SA Pacific Standard Time:20200714T160000
CLASS:PUBLIC
PRIORITY:5
DTSTAMP:20200713T222111Z
TRANSP:OPAQUE
STATUS:CONFIRMED
SEQUENCE:0
LOCATION:Reunión de Microsoft Teams
```

DTSTART and DTEND has not "Z" suffix so that indicates that those datetime are not provided in UTC.  So I think that you don't need translate the timezone:
https://tools.ietf.org/html/rfc5545#section-3.2.19

**If applicable, add screenshots or screen recordings to help explain your problem.**

I'm using maya-calendar 5.0.6+r1806+pkg130~ubuntu5.1.5.1 and evolution too (it's fine on this). You can look the differences.
87380020-33cb2780-c557-11ea-907f-bfcc77ee7a2e
87359305-25681600-c52d-11ea-9bef-160165dbbd3e

**### To Reproduce**
You can import ICS calendar without Z suffix in DTSTART and DTEND in Maya Calendar

**## Platform Information**
OS: elementary OS 5.1.6 Hera x86_64
Kernel: 5.3.0-62-generic

Related bugs: 
#92 
#50

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92263814-windows-timezones-are-not-handled?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60388692&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mcclurgm:
Fixes #572, possibly others

As discussed in #660, the conversion process from libical to GLib timezones is very brittle. The most visible result is that Windows-style timezones (`Western European Standard Time`) are not handled.

This PR moves all timezone logic to libical, which should have access to all the necessary information to run timezone conversions properly, regardless of whether the timezone corresponds to a system built-in. I accomplished this by modifying the functions that convert `ICal.Time` and `ICal.Component` to `GLib.DateTime`. 

Basically, the change in this PR is converting the timezone in libical before calling `icaltime_to_datetime`, instead of converting after in GLib. That's the only subtantive change, really.

One restriction is that some of thsee convert to DateTimes in non-local timezones. Since it's not always possible to get a `GLib.TimeZone` for these cases, the timezone in resulting DateTimes isn't guaranteed to be correct. I documented this restriction with emphasis, but it's still not ideal.

I added lots of tests to make sure this works as intended. So I'm pretty sure the new/modified functions all work properly, but the question is if there are any clients that need to be fixed. A text search for function calls didn't reveal anything, but in review I'd appreciate some help looking. If there are no weird cases I missed, then the modified functions should be a drop-in replacement.


danrabbit:
Thanks for all of your work on this!

Issue #694, opened 2021-08-21
Deprecation Warning: Hdy.Avatar.set_image_load_func


marbetschar:
While compiling the latest master, I noticed the following warning:

```
../src/EventEdition/GuestGrid.vala:111.29-111.54: warning: `Hdy.Avatar.set_image_load_func' has been deprecated since 1.2
```

I had the same warning in elementary Mail and fixed it using [`Hdy.Avatar.set_loadable_icon`](https://github.com/elementary/mail/blob/9477c24201d3c9c3c7bcc650e0290daa0ee1ab21/src/MessageList/MessageListItem.vala#L293). Furthermore I've seen [`Folks`'s avatar seems to be a `LoadableIcon`](https://valadoc.org/folks/Folks.AvatarDetails.avatar.html) - so it might be possible to simply do the following to fix the warning?

```vala
avatar.set_loadable_icon (individual.avatar);
```

danrabbit:
Looks like this was fixed in #716

PR #716, opened 2021-10-13
694: Replace deprecated function call. Add null check.

jpgleeson:
Update deprecated function call and gate behind null check. #694


jeremypw:
The code appears to make sense but not sure how to check.  An existing event linked to my Google account did not show my Google avatar (just the fallback first letter one).

jeremypw:
There is some extraneous white space at the end of line 111 causing lint failure.

jpgleeson:
Linting fixed in 14acae1.

I believe it won't show the correct remote avatar until we set up a backend to fetch and store these avatars to the local machine similar to the Mail backend for Local and Gravatar. I've been tipping away at a PR for this but it won't be for some time.

jeremypw (approved):
LGTM -thanks!

PR #717, opened 2021-10-19
Application: Force use of portals

danirabbit:
Since Calendar currently can't be sandboxed due to online accounts, this forces the use of portals even outside of flatpak. What this means practically is that Calendar uses the native filechooser instead of the one from Gtk

I didn't notice any issues with reading system gsettings like the show weeks setting and location services seem to work as normal


jeremypw:
Code is already setting GTK_USE_PORTAL.

zeebok (approved):
The code seems alright. Dani seems very sure this works, so I haven't tested. Let me know if I should and what exactly I should look for.

src/Application.vala
```diff
@@ -52,6 +52,7 @@ namespace Maya {
 
             application_id = Build.EXEC_NAME;
 
+            Environment.set_variable ("GTK_USE_PORTAL", "1", true);
```
Marukesu, line 55: I think the same comment that mail has about "removing this when Gtk4/Gtk.FileDialog is available" should be here too.

Marukesu:
About the notification working when it wasn't before, that's because GLib/Gio store the environment value during the library ctor, while Gtk only do it after calling Gtk.init(). so GLib code isn't affected by this.

src/Application.vala
zeebok, line 55:
```suggestion
            // FIXME: Remove once ported to Gtk.FileDialog
            Environment.set_variable ("GTK_USE_PORTAL", "1", true);
```

zeebok:
```suggestion
            // FIXME: Remove once ported to Gtk.FileDialog
            Environment.set_variable ("GTK_USE_PORTAL", "1", true);
```

PR #722, opened 2021-11-06
Revert "utils: Workaround wrong ICal.Time.get_timezone() binding to fix memleak (#719)"

tintou:
Also add code to release the internal cache at the end of the program

### DIFF
--- a/core/Services/Calendar/Util/ICalTime.vala
+++ b/core/Services/Calendar/Util/ICalTime.vala
@@ -33,57 +33,33 @@ namespace Calendar.Util {
             return new GLib.TimeZone.local ();
         }
 
-        var tzid = date.get_tzid ();
+        unowned string? tzid = date.get_tzid ();
         if (tzid == null) {
             // In libical, null tzid means floating time
             assert (date.get_timezone () == null);
             return new GLib.TimeZone.local ();
         }
 
-        // Otherwise, get timezone from ICal
-        // First, try using the tzid property
-        if (tzid != null) {
-            /* Standard city names are usable directly by GLib, so we can bypass
-             * the ICal scaffolding completely and just return a new
-             * GLib.TimeZone here. This method also preserves all the timezone
-             * information, like going in/out of daylight savings, which parsing
-             * from UTC offset does not.
-             * Note, this can't recover from failure, since GLib.TimeZone
-             * constructor doesn't communicate failure information. This block
-             * will always return a GLib.TimeZone, which will be UTC if parsing
-             * fails for some reason.
-             */
-            var prefix = "/freeassociation.sourceforge.net/";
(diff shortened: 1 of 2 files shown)


core/Services/Calendar/Util/ICalTime.vala
```diff
@@ -62,22 +64,25 @@ namespace Calendar.Util {
                 return new GLib.TimeZone (tzid);
             }
         }
-
-        //FIXME See https://github.com/libical/libical/pull/513
         // If tzid fails, try ICal.Time.get_timezone ()
-        ICal.Timezone* timezone = date.get_timezone ();
+        unowned ICal.Timezone? timezone = null;
+        if (timezone == null && date.get_timezone () != null) {
```
jeremypw, line 69: Why test for `null` here when the previous line set it to `null`?

tintou: Hmm, I did a blind reverse of the commit but it fixed some things too

jeremypw:
@mcclurgm Could you just check this as you know more about timezone and ICal than I do?

mcclurgm:
This is really exposing some more problems we have in the `icaltime_get_timezone` function. The unreachable code you deleted would be necessary to account for failed timezone creations (like anything that doesn't have the Olson TZID format), but glib only ever fails silently so I never accounted for that. Some tech debt that was my fault. I think it would be a good thing to add it back at some point, unless we can properly get rid of GLib timezones altogether, since if there's one thing I've learned trying to deal with converting back and forth it's that the whole process is messy and full of holes, and best avoided.

Otherwise, it seems like the original #719 was accounting for the memory issue the wrong way, and the proper way is to free libical's global timezone cache when the app exits. Is that correct? I'm not familiar with Vala manual memory management, but it looks like the old PR was improperly adding a free call, which I suppose makes sense to remove. 

Otherwise, the changes made to `icaltime_get_timezone` make sense given what I said above. I'll trust you about `free_global_objects` I suppose. So based on what I can understand, this makes sense to me.

jeremypw:
@mcclurgm So you would not anticipate any serious regressions from this? I am not sure how to test the timezone stuff.

mcclurgm:
Compared to the state before #719, I don't expect any regressions. The changes to `icaltime_get_timezone` should have no effect (any removed code was either duplicate or unreachable), and otherwise the only change is adding the call to `free_global_objects`. I expect no timezone regressions. We also have tests that should indicate regressions if I wrote them right.

Compared to the state after #719, that's far enough outside my knowledge area I don't know. Hopefully that's enough info for you.

jeremypw:
Thanks @mcclurgm, I think I can approve on that basis.

jeremypw (approved):

Issue #553, opened 2020-06-25
The behaviour when creating new events should be logical and consistent with regard to the "mark as default calendar" option


4jNsY6fCVqZv:
![Screen_record cleaned](https://user-images.githubusercontent.com/46536067/85727370-792dcf00-b6e6-11ea-830e-b3832a0e81da.gif)

One thing irritates me:

At first glance, it seems that it is mandatory to define a default calendar in order to create an event. However, test5 & test6 can be created after I have already unchecked the default calendar option again. That seems inconsistent to me.

Beyond my observation, I wonder whether it should be necessary for the creation of new events to have at least one standard calendar defined before. And I also wonder if it should even be possible to start creating an event using the "Create an event" button without setting a default calendar. Or whether the button should remain insensitive until I have done so.

My expectation as a user would in any case be consistent behaviour.

I use the latest stable and unmodified Hera version including all updates.

Somehow also connected with this issue https://github.com/elementary/calendar/issues/544
Again, the "Create Event" dialog can be started, although it looks like the conditions for creating a new event are not yet fulfilled.

mcclurgm:
I have a feeling this is caused by #550, but I'm not sure. Since the check/uncheck behavior doesn't seem to work at all, I wouldn't be surprised if there actually is still a default calendar but the check box doesn't reflect that. This is all speculation though.

4jNsY6fCVqZv:
Do you have an idea how we can find out and solve this?

mcclurgm:
No sorry... I've never dealt with that part of the codebase. I imagine that the signal we send when you click the check button may be wrong. If you have experience looking at our code, that's where I'd start, but I'm going to (work allowing) do that soon too.

mcclurgm:
We just fixed #550, so the checkbox behavior should be at least a little nicer. I don't really understand what the issue is here, so can you try rebuilding from git and let me know if you're still seeing it?

4jNsY6fCVqZv:
Unfortunately I don't have time to rebuild it from git right now. Maybe someone else?

mcclurgm:
Fair enough. We just released an update with the changes, so if you have the time to give that a try that would be great! Otherwise no worries.

mcclurgm:
@4jNsY6fCVqZv Now that Odin is released, could you have a look and see if the issue still exists? I don't really understand what the issue is you're describing here, but the current release has the fix to the default calendar picker.

4jNsY6fCVqZv:
[...]

PR #730, opened 2021-12-04
Make default calendar checkbox insensitive if read-only

mcclurgm:
Fixes an issue reported in #553. Previously we didn't check whether a calendar is read-only, so you could set it as default, with some strange results. This fixes that by disabling the checkbox.

Since a disabled checkbox can now mean multiple things (read-only calendar if unchecked, or already default if checked), should we add a label explaining why it's disabled? I'm not sure if that's necessary, but it could be helpful.

### DIFF
--- a/src/SourceDialog/SourceDialog.vala
+++ b/src/SourceDialog/SourceDialog.vala
@@ -270,8 +270,9 @@ public class Maya.View.SourceDialog : Gtk.Grid {
             try {
                 var registry = new E.SourceRegistry.sync (null);
                 var source_is_default = source.equal (registry.default_calendar);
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
This has the expected effect and makes sense (although I am not entirely clear why my Births and Anniversaries calendar is read-only?)

mcclurgm:
My assumption is that it polls that information from contact details, although that's speculative as I don't use it.

jeremypw:
@mcclurgm  Yes, I think your right. I am not sure how to add any birthdays and anniversaries using other Elementary apps. Friends is not released yet and doesnt seem to handle birthdays and anniversaries.

PR #733, opened 2021-12-18
Application: Add `Calendar.Settings` Singleton 

Marukesu:
Second part of elementary/settings-daemon#39.

This create an new `Calendar.Settings` class to provide both internal and external settings, while allowing calendar to work ever if they can't access any of the settings (eg. running `./build/src/io.elementary.calendar`).

this is exposing two unused keys right now:
 * background for #681
 * first_weekday for #638

`Maya.Settings` was removed and `time_format ()` moved there.


jeremypw (commented):
Just a few comments.

The handling of WindowState seems a bit obscure - is this so it works with the portal?

zeebok:
Some conflicts with main to sort through :)

Marukesu:
Well, with the bakground portal not needing this anymore, and with the first_weekday PR closed, i think this PR lost purpose, for now.

unless the ability to run without the schemas installed is a much wanted feature, what i doubt it is.

zeebok:
That really only seems useful for some dev testing scenarios. Granted this was given to the OS 7.1 project and is part of the upcoming release draft so maybe @danirabbit can confirm if this should stay or not.

danirabbit:
Ah yeah i think this was on the project because at the time it was background portal related maybe? We can bump it off it doesn’t seem like it’s currently useful

zeebok:
Based on the last few messages it sounds like this can be closed.

PR #737, opened 2021-12-27
Application: Make the option fields internal to the Application class

tintou:
This fixes a compilation error with latest valac which doesn't allow the option
variables to be less accessible than the OptionEntry.


danrabbit (approved):
Makes sense :+1:

PR #739, opened 2022-01-18
Update appdata.xml for delete dialog

mcclurgm:
I would appreciate if you have better wording than this; I feel like what I wrote is not the best.

I don't know what our next release would be, so I assumed it would be a .0.1 update.


danrabbit:
6.1.0 wasn't release yet, so moved the line to that release :)

danrabbit (approved):

PR #747, opened 2022-03-16
Use assert_true instead of assert in tests

mcclurgm:
What it says on the tin. For unit tests, you're supposed to use `GLib.assert_true` instead of `GLib.assert`, which may be disabled by compilation flags. See https://docs.gtk.org/glib/testing.html.


mcclurgm:
Hmm, this CI action really doesn't want to run. Do you have any idea why it wouldn't even show up in the Actions tab? I'm not familiar with Github actions, so I'm not sure where to look.

davidmhewitt:
CI probably didn't run due to the issues GitHub has been experiencing in the last couple of days. I've just pushed a completely blank commit, to trigger it. Looks to be running now

mcclurgm:
Thanks! Didn't think of that 😅️

PR #758, opened 2022-08-07
Add support for Evolution Data Server 3.46

tintou:
EDS 3.46 has switched to libsoup-3


meisenzahl (approved):
I can confirm that this builds on Fedora 37

Conan-Kudo:
cc: @decathorpe

PR #760, opened 2022-09-04
Add check for termination of by_day_array

### LINKED ISSUE #759: Editing event automatically enables Saturday repeat
### What Happened?

I edited an event I already made that repeats on Monday and Wednesday. I made a change to the first event, and when looking at the Repeat tab, I found Saturday was selected, even though it was never selected in the first place.

### Steps to Reproduce

1. Create a repeating event on any weekday that isn't Saturday
2. Edit the repeating event
3. Look in the Repeat tab to find Sat is selected

### Expected Behavior

I should have to explicitly click on Sat for it to repeat on Saturday, rather than this happen automatically

### OS Version

Other Linux

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

OS: Fedora Linux 36 (Workstation Edition)
Kernel: 5.18.16-200.fc36.x86_64
DE: GNOME 42.4
WM: Mutter
CPU: 11th Gen Intel i7-1165G7 (8) @ 4.700GHz
GPU: Intel TigerLake-LP GT2 [Iris Xe Graphics]
Memory: 16GB

mcclurgm:
Fixes #759 

Description of the reasoning there. The array has a special terminator value, which I added a check for. If it's encountered, stop looping over the array, since further entries are no longer valid.

### DIFF
--- a/src/EventEdition/RepeatPanel.vala
+++ b/src/EventEdition/RepeatPanel.vala
@@ -273,7 +273,11 @@ public class Maya.View.EventEdition.RepeatPanel : Gtk.Grid {
     private void load_weekly_recurrence (ICal.Recurrence rrule) {
         repeat_combobox.active = 1;
         var by_day = rrule.get_by_day_array ();
-        for (uint i = 0; i < by_day.length; i++) {
+        for (
+            uint i = 0;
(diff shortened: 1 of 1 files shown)


tintou (approved):
Good catch

PR #762, opened 2022-10-01
Fix meson warning for run_command

turbobastii:
Since I am already fixing the other repos with this issue, I thought I might do it here aswell.

TLDR: meson will change the default behaviour of `run_command` and displays that as a warning when building.
See: https://github.com/mesonbuild/meson/issues/9300

image

### DIFF
--- a/core/Tests/meson.build
+++ b/core/Tests/meson.build
@@ -19,7 +19,7 @@ util_tests = executable(
 test('Utils Tests', util_tests, suite: 'core', is_parallel: false)
 
 # We need these three locales to run the tests
-locales = run_command('locale', '-a').stdout().split('\n')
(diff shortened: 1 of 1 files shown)


tintou:
Don't we want it to be true?

turbobastii:
Well it was false implicitly by default, but if you think it would make more sense if it was true, I can of course change it.

eli-schwartz:
> Well it was false implicitly by default

If the implicit default was generally good, then we (Meson) would not have deprecated it in the hope of eventually changing the default. :)

When adding an explicit value, it is wisest to consider what that value would be best served by being, rather than just explicating the default purely to quell the warning.

turbobastii:
Very well put, thank you. I changed it accordingly.

lenemter (approved):

Issue #761, opened 2022-09-29
Support glib 2.73.1+: Avoid property notification during object construction


bobby285271:
### What Happened?

<img src="https://user-images.githubusercontent.com/20080233/190838919-782843ef-7700-481e-bf7c-738572c4c964.png" width=50%>

In https://github.com/elementary/calendar/issues/756#issuecomment-1249990402, we noticed that the calendar app does not show any date numbers, after making a minimal reproducer in https://github.com/bobby285271/e-calendar-issue and bisecting glib commits (thanks `@jtojnar`!), we noticed that [this glib commit](https://gitlab.gnome.org/GNOME/glib/-/commit/8851112b8de024ff6fb4235a4475ffa6c98973dc) intentionally changes the behavior, which is part of glib 2.73.1+. It will be nice to adjust the code here so the app can still work nicely with glib 2.73.1+.

The affected code is at

https://github.com/elementary/calendar/blob/7e65ae0053696b5dae1453ec3ceac391ed870890/src/Grid/GridDay.vala#L104-L107

The workaround is at https://github.com/elementary/calendar/issues/756#issuecomment-1252400047.

### Steps to Reproduce

1. Build with glib 2.73.1+
2. Launch the app
3. Notice the app does not show any date numbers

bobby285271:
I actually don't know how bad the affect will be for other Pantheon apps, for example in https://github.com/NixOS/nixpkgs/pull/192686#issuecomment-1256840424 we also noticed that iconbrowser is affected.

jtojnar:
[This handler](https://github.com/elementary/calendar/blob/641abec284a39072ac4c3a11fdcaa1ce71ad5fa4/core/Services/Calendar/EventStore.vala#L87) might be suffering from the same issue. Although `compute_ranges()` is [already called](https://github.com/elementary/calendar/blob/641abec284a39072ac4c3a11fdcaa1ce71ad5fa4/core/Services/Calendar/EventStore.vala#L80) in the constructor, the [other two methods](https://github.com/elementary/calendar/blob/641abec284a39072ac4c3a11fdcaa1ce71ad5fa4/core/Services/Calendar/EventStore.vala#L453-L454) are not. But I have not checked if they are necessary and am not sure how to test the EventStore.

PR #763, opened 2022-10-01
Fix day in month in Grid with GLib 2.73.1+

jtojnar:
GLib 2.73.1 [no longer fires notify signal for properties set during instantiation](https://gitlab.gnome.org/GNOME/glib/-/commit/8851112b8de024ff6fb4235a4475ffa6c98973dc). This resulted in the label remaining empty. Now, the `Grid.update_day` would also run and set `date` property but that did not have an effect – probably because the date did not change.

We need to set the initial label ourselves. To avoid code duplication, we can just use `GBinding`.

Fixes: https://github.com/elementary/calendar/issues/761

cc @bobby285271

### DIFF
--- a/src/Grid/GridDay.vala
+++ b/src/Grid/GridDay.vala
@@ -101,8 +101,10 @@ public class Maya.View.GridDay : Gtk.EventBox {
         Gtk.TargetEntry dnd = {"binary/calendar", 0, 0};
         Gtk.drag_dest_set (this, Gtk.DestDefaults.MOTION, {dnd}, Gdk.DragAction.MOVE);
 
-        this.notify["date"].connect (() => {
-            label.label = date.get_day_of_month ().to_string ();
+        this.bind_property ("date", label, "label", BindingFlags.SYNC_CREATE, (binding, srcval, ref targetval) => {
+            unowned var date = (GLib.DateTime) srcval.get_boxed ();
+            targetval.take_string (date.get_day_of_month ().to_string ());
+            return true;
         });
     }
 


src/Grid/GridDay.vala
```diff
@@ -101,8 +101,10 @@ public class Maya.View.GridDay : Gtk.EventBox {
         Gtk.TargetEntry dnd = {"binary/calendar", 0, 0};
         Gtk.drag_dest_set (this, Gtk.DestDefaults.MOTION, {dnd}, Gdk.DragAction.MOVE);
 
-        this.notify["date"].connect (() => {
-            label.label = date.get_day_of_month ().to_string ();
+        this.bind_property ("date", label, "label", BindingFlags.SYNC_CREATE, (binding, srcval, ref targetval) => {
+            DateTime date = (DateTime) srcval;
```
tintou, line 105:
```suggestion
            unowned var date = (GLib.DateTime) srcval.get_boxed ();
```
I know that the GValue cast is automatic in Vala but I find explicit getter safer here

src/Grid/GridDay.vala
```diff
@@ -101,8 +101,10 @@ public class Maya.View.GridDay : Gtk.EventBox {
...
+            targetval.set_string (date.get_day_of_month ().to_string ());
```
tintou, line 106:
```suggestion
            targetval.take_string (date.get_day_of_month ().to_string ());
```

jtojnar:
Thanks, applied the suggestions, seems to still work fine.

tintou (approved):

Issue #668, opened 2021-05-18
Incorrect accent color in light and dark mode


greenmanspirit:
## What Happened
When I click on a day or an event in the sidebar, the day number circle and event background are lime when my accent color is grape. I tried different accent colors, both light and dark modes and logging out and back in. In all cases, the color is lime.

## Expected Behavior
The color of the day number circle and event background to be grape.

## Steps to Reproduce

1. Set accent color to grape
2. Open Calendar
3. Select days and events

### Logs

No logs

## Platform Information

elementary OS 6.0 Beta fully patched from repo.

tintou:
I think that this is the intended behavior as the application itself is using the lime color accent. Asking @elementary/ux to confirm the intent.

cassidyjames:
Yeah, as noted in System Settings, 

>Used across the system by default. **Apps can always use their own accent color.**

That's what's happening here.

PR #780, opened 2023-04-30
Style updates

### LINKED ISSUE #779: Scrolling happens vertically but buttons indicate paging horizontally
### What Happened?

When the calender app is opened and one hovers the cursor over the grid of days and scrolls, it scrolls vertically. But the buttons seem to indicate horizontal paging. The top bar calender works by paging horizontally, smoothly with multitouch too.

### Steps to Reproduce

1. open the calendar app
2. hover the date grid
3. scroll

### Expected Behavior

I would expect horizontal scrolling instead, like in the top bar calender widget.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
- Use system accent color instead of a custom one. This was discussed previously in #668. I personally think there is no reason to use a custom accent color (lime green) here but I suppose that's personal preference.
- Slide months right/left instead up/down (Closes #779). Also up to personal preference but I think it makes sense to stay consistent with the arrows in the headerbar and with how the wingpanel indicator handles it.

These changes are a **suggestion** please reject if they don't make sense designwise :)


zeebok (dismissed):
Code-wise this looks fine. Definitely agree that UX needs to weigh in on what they think

zeebok:
@danirabbit can we consider you adding this to the release as a thumbs up to using user accent color?

data/style/Application.css
zeebok, line 1: Since this empties the file, it might be good to remove the file and update the Application.vala and gresource to not include it

leolost2605: Done

zeebok (approved):
Thanks! Looks good to me now and confirmed to work :)

PR #786, opened 2023-05-25
Meson option to build without libchamplain

meisenzahl:
Using `libsoup2` and `libsoup3` in the same process is not supported. On distributions with more recent dependencies, `libchamplain` still uses `libsoup2` as dependency. Therefore, Calendar crashes immediately after startup.

```sh
$ io.elementary.calendar 

(process:1855): libsoup-ERROR **: 06:31:48.760: libsoup2 symbols detected. Using libsoup2 and libsoup3 in the same process is not supported.
Trace/breakpoint trap (core dumped)
```

This PR adds a flag to disable libchamplain as a dependency. Tested on Fedora Rawhide.

## With `libchamplain` 

Bildschirmfoto von 2023-05-25 06 36 00

## Without `libchamplain`

Screenshot from 2023-05-25 06 37 01

@tintou has provided a patch for libchamplain [that allows to use libchamplain with libsoup3 libraries and applications](https://gitlab.gnome.org/GNOME/libchamplain/-/merge_requests/13). If a distribution cannot use this patch, we can provide an alternative with this PR.

### DIFF
--- a/src/EventEdition/EventDialog.vala
+++ b/src/EventEdition/EventDialog.vala
@@ -42,7 +42,7 @@ public class EventDialog : Granite.Dialog {
 
         private EventEdition.GuestsPanel guests_panel;
         private EventEdition.InfoPanel info_panel;
-#ifdef USE_LIBCHAMPLAIN
+#if USE_LIBCHAMPLAIN
         private EventEdition.LocationPanel location_panel;
 #endif        
         private EventEdition.ReminderPanel reminder_panel;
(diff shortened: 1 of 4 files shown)


src/EventEdition/EventDialog.vala
```diff
@@ -42,7 +42,9 @@ public class EventDialog : Granite.Dialog {
 
         private EventEdition.GuestsPanel guests_panel;
         private EventEdition.InfoPanel info_panel;
+#if USE_LIBCHAMPLAIN
         private EventEdition.LocationPanel location_panel;
+#endif
```
meisenzahl, line 47:
```suggestion
#endif
```

PR #787, opened 2023-05-25
Allow to fallback to statically linked libchamplain

tintou:
This allows to solve the libsoup2/3 issue in some distributions.


meisenzahl (approved):
I can confirm that this works smoothly on Fedora Rawhide!

PR #793, opened 2023-08-02
Create maya-calendar.postinst

danirabbit:


lenemter (approved):

lenemter:
Wait, can it be named not `postinst`?

danirabbit:
@lenemter no because there are multiple packages created here, so we need to prefix with the package name so that debbuild knows which of these packages this postinst script belongs to

lenemter:
@danirabbit Oh ok, just cheked in [Debian manual](https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html#maintscripts) and it should work.

PR #798, opened 2023-09-18
Switch to flat headerbars

leolost2605:
Brings Calendar in line with Mail, Videos etc.
I'm not quite sure about the button order or the removal of the calendar grid header styling etc. so any suggestions are welcome. Or, if this doesn't fit Calendar altogether, just close this PR.

Screenshot:
Screenshot from 2023-09-18 20 59 36


danirabbit (approved):
I think this is a great start on modernizing the design here! :rocket:

PR #803, opened 2023-11-10
Rearrange headerbar items

danirabbit:
Re-arranges some header items so that they are more closely associated with the pane they directly act on:

* The calendar settings and export buttons move to the calendar pane since the export button acts on the default calendar, not the currently visible agenda items, and the settings popover affects which calendars are shown in the calendar view
* The create new event button is now in the agenda pane since the event will be created for the selected day as labeled in this pane
* The agenda pane's label moves up into the header to make more room for events

## BEFORE
Screenshot from 2023-11-10 11 15 42

## AFTER
Screenshot from 2023-11-10 11 21 26

### DIFF
--- a/src/AgendaView.vala
+++ b/src/AgendaView.vala
@@ -31,39 +31,35 @@ public class Maya.View.AgendaView : Gtk.Box {
     private DateTime selected_date;
 
     construct {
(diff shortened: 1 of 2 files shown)


zeebok (approved):
Looks good! Both the code and the new design!

PR #820, opened 2025-02-24
chore(flatpak): bump EDS version

davidmhewitt:
Haven't tested the resulting flatpak, but this matches https://github.com/elementary/mail/pull/1012

PR #821, opened 2025-02-24
CI: run arm workflows on arm runner

davidmhewitt:


ryonakano (commented):
LGTM, but maybe we should wait merging this until #820 is merged because this updates the container image from 7.1 to 8

PR #829, opened 2025-03-01
InfoPanel: absorb CalendarButton

danirabbit:
* Can't subclass Gtk.MenuButton in GTK4
* code style
* replace grid with box
* Use a property binding for source and `SYNC_CREATE`


src/EventEdition/InfoPanel.vala
```diff
@@ -185,6 +208,17 @@ public class Maya.View.EventEdition.InfoPanel : Gtk.Grid {
         attach (frame, 0, 9, 5, 1);
 
         load ();
+
+        calchooser.bind_property ("current-source", current_calendar_grid, "source", SYNC_CREATE);
+
+        calchooser.notify["current-source"].connect ((s, p) => {
+            calendar_button.tooltip_text = "%s—%s".printf (current_calendar_grid.label, current_calendar_grid.location);
```
jeremypw, line 215: The tooltip text does not get set unless current-source changes? This occurs also in `master` though.

jeremypw (approved):
This reproduces the behaviour of the `master` branch faithfully and the code looks good.  

There is a pre-existing issue with the button tooltip and personally I would prefer to see the calendar the event is associated with even if there is only one writable.

src/EventEdition/InfoPanel.vala
danirabbit: Yeah I noticed that too. I think we'll revisit this anyways when porting to Gtk.DropDown, but like you said already occurs in master so wasn't too invested in try to fix it in this branch

PR #830, opened 2025-03-03
Bump to platform 8.2

danirabbit:


ryonakano:
I fixed Calendar builds and runs successfully with Flatpak both on amd64 and arm64. Also bumped platform from 8.1 to 8.2.

PR #840, opened 2025-07-21
SourceDialog: make a dialog

danirabbit:
It looks like this was a dialog and then at some point became a popover subpage? Forms should definitely be a new dialog and not a popover since popovers can be easily/accidentally dismissed

Screenshot from 2025-07-21 14 20 08


danirabbit:
Yeah spacing is terrible here. Gonna get it in a follow up. Thanks for the review!

PR #854, opened 2025-07-23
EventButton: don't subclass Revealer

danirabbit:
* Bump copyright header
* Revealer is sealed in GTK4, so don't subclass revealer
* Move hide/show functions internally so we don't have to make the revealer publicly accessible
* Minor code style while we're here

### DIFF
--- a/po/ca.po
+++ b/po/ca.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: maya\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2026-01-28 15:27+0000\n"
-"PO-Revision-Date: 2025-11-22 12:55+0000\n"
+"PO-Revision-Date: 2026-01-29 18:55+0000\n"
 "Last-Translator: David M <rbndavid@gmail.com>\n"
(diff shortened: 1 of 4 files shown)


jeremypw (changes requested):
The hide/reveal mechanism does not work correctly. The events hide but do not re-appear when space is available.

danirabbit:
@jeremypw can you let me know how to reproduce the issue? I was testing by enabling and disabling calendars

jeremypw:
@danirabbit I was testing by resizing the calendar window.  In master the event buttons  are replaced by a placeholder below a certain size and re-appear on enlarging the window again.  In this PR the buttons disappear but do not reappear (unless the calendar is scrolled to another month and back.)

danirabbit:
@jeremypw ah I misunderstood what this was for. Thank you! It should be fixed now

jeremypw (changes requested):
Latest commits fix the reveal issue previously mentioned.
I found another issue where appearance differs from main. If you add a second event to a day where where there is only just room to display  one event you get overlapping text:
Screenshot from 2025-07-30 12 57 57

In main the event is replaced by "2 more..."

PR #930, opened 2026-02-18
VAutoHider: keep a list of children

danirabbit:
There's no `get_children` in GTK4, so keep our own list of children because `observe_children` is supposedly expensive


zeebok (approved):

ryonakano:
@danirabbit I resolved conflicts here. Is there any reason this is in draft? Do we want to apply suggestion from @lenemter?

ryonakano (approved):
Approving as LGTM.

(I personally think @lenemter's suggestion depends on preferred coding style, so I'm fine either we apply it or not)

danirabbit:
I guess I don't know that this really solves anything. I hadn't gotten to a solution in GTK4 that was actually working so wasn't sure if this was the right direction

PR #932, opened 2026-02-26
LocationPanel: remove point dragging to set location

### LINKED ISSUE #155: After dragging location pointer, map zoom level should adjust appropriately.
When dragging the location a significant distance (map zoomed out), it would be convenient if the when the pointer is dropped, the map is automatically zoomed in to show the area around the pointer in more detail (proportionate to the level of accuracy of placement of the pointer in the zoomed out view), so that the pointer may be adjusted more precisely.

danirabbit:
Closes #155 

This doesn't feel like a reasonable way to set locations tbh. Neither iOS Calendar, Fastmail Calendar, nor Google Calendar allow this


zeebok (approved):
I didn't even know there was a way to click and drag the map marker! I do wish there was some kind of indication that the address I am typing is correctly being found, or on-the-fly search results or something

PR #933, opened 2026-02-26
EventButton: Clean up drag n drop events

danirabbit:
Just some minor refactoring for DND. Uses a function instead of a lambda, sets selection data based on target type


jeremypw (approved):
Looks good.  Moving events between days and dropping onto e.g. Files works as before.

I note you cannot drop an event onto Code but that is also true for `master`.  It would be possible to add a "text" target.

I also note that there is no drag icon but maybe that was always the case.  It would be better if there were one but that can be left for another PR.

PR #939, opened 2026-04-09
VAutoHider: use ListBox

danirabbit:
This gives us sorting and solves some of our issues with keeping track of children


ryonakano:
@danirabbit Are there any TODOs for this PR to be ready for review?

danirabbit:
I can't remember what the state of it was except that it wasn't working properly 😅

PR #940, opened 2026-04-09
VAutoHider: cleanup size_allocate

danirabbit:
* Avoid nesting and using `continue` where possible
* Brackets with if statements
* DRY `more_label` hide and vexpand
* rename `height` to `shown_children_height` for clarity

### DIFF
--- a/src/Grid/VAutoHider.vala
+++ b/src/Grid/VAutoHider.vala
@@ -49,57 +49,43 @@ public class Maya.View.VAutoHider : Gtk.Bin {
 
     public override void size_allocate (Gtk.Allocation allocation) {
         base.size_allocate (allocation);
-        int global_height = allocation.height;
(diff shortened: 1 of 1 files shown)


jeremypw (changes requested):
This seems to alter the behaviour when there is only one event. The `more_label` appears when there is less space than the child height + `more_label_height` whereas in `master` it appears when there is less space available than the child_height.  So the `more_label` appears earlier than necessary when shrinking the height.

It would simplify matters if we can assume/ensure the child height is constant and the same as the `more_label` height.

danirabbit:
@jeremypw ah thanks for noticing that! I think it should be fixed now. The logic here was pretty hard to follow. Hopefully it's a little clearer now

jeremypw (approved):
Works as expected now, thanks.

PR #953, opened 2026-08-11
Icons: use red binding

danirabbit:
Following up on https://github.com/elementary/icons/pull/1419 and some discussion in #ux on Discord, I think we generally agreed on making the calendar red


wpkelso (approved):
Let's go 🚀 

We should consider if the default dock ordering needs to be revisited. If I remember correctly it's been ordered by color in the past, but that doesn't seem held to now.

